PR #7, opened 2018-08-27
adds new client library

catherinejones:
This change adds a simple library for interacting with a Voucher server.

### DIFF
--- a/client/signimage.go
+++ b/client/signimage.go
@@ -1,0 +1,27 @@
+package client
+
+import (
+	"time"
+
+	"github.com/Shopify/voucher"
+	"github.com/docker/distribution/reference"
+)
+
+const timeout = 120 * time.Second
+
+// SignImage takes an image URL (which includes the registry and the image
+// digest) as well as the check to be performed and makes a call to Voucher
+// to run the specified checks. It returns an error, or nil if no errors
+// occur and the check was successful.
+func SignImage(hostname string, image reference.Canonical, check string) (voucher.Response, error) {
+	client, err := NewClient(hostname, timeout)
+	if nil != err {
+		return voucher.Response{}, err
+	}
+
+	checkResp, err := client.Check(check, image)
+	if nil != err {
+		return voucher.Response{}, err
+	}
+	return checkResp, nil
+}
--- a/cmd/voucher_client/attest.go
+++ b/cmd/voucher_client/attest.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/Shopify/voucher"
+	"github.com/Shopify/voucher/auth/google"
+	"github.com/Shopify/voucher/client"
+	"github.com/docker/distribution/reference"
+)
+
+const timeout = 120 * time.Second
+
+func lookupAndAttest(hostname, check, image string) error {
+	var ok bool
+	var namedRef reference.Named
+
+	ref, err := reference.Parse(image)
+	if nil != err {
+		return fmt.Errorf("parsing image reference failed: %s", err)
+	}
+
+	if namedRef, ok = ref.(reference.Named); !ok {
+		return fmt.Errorf("couldn't get named version of reference: %s", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	voucherClient, err := voucher.AuthToClient(ctx, google.NewAuth(), namedRef)
+	if nil != err {
+		return fmt.Errorf("creating authenticated client failed: %s", err)
+	}
+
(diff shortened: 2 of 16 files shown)


client/signimage.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/Shopify/voucher"
+	"github.com/docker/distribution/reference"
+)
+
+// SignImage takes an image URL (which includes the registry and the image
+// digest) as well as the check to be performed and makes a call to Voucher
+// to run the specified checks. It returns an error, or nil if no errors
+// occur and the check was successful.
+func SignImage(hostname string, image reference.Canonical, check string) (voucher.Response, error) {
+	client, err := NewClient(hostname, 60*time.Second)
```
klautcomputing: Will 60 seconds be enough for a big image, that we have never scanned before?

cmd/voucher_client/attest.go
```diff
@@ -0,0 +1,50 @@
...
+func lookupAndAttest(hostname, check, image string) error {
+	var ok bool
+	var namedRef reference.Named
+
+	ref, err := reference.Parse(image)
+	if nil != err {
+		return fmt.Errorf("parsing image reference failed: %s", err)
+	}
+
+	gcloudToken, err := voucher.GetAccessToken()
```
klautcomputing: Let's make sure that we update this w/r/t the other open PR

cmd/voucher_client/attest.go
```diff
@@ -0,0 +1,50 @@
...
+	if nil != err {
+		return fmt.Errorf("authorizing failed: %s", err)
+	}
+
+	canonicalRef, err := getCanonicalReference(oauth, namedRef)
+	if nil != err {
+		return fmt.Errorf("getting image digest failed: %s", err)
+	}
+
+	fmt.Printf(" - Attesting image: %s\n", canonicalRef.String())
```
klautcomputing: Why are we using `fmt` here instead of `log`?

cmd/voucher_client/digest.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+
+	"github.com/Shopify/voucher/docker"
+	"github.com/docker/distribution/reference"
+)
+
+// getCanonicalReference gets the canonical image for the passed reference.
+// If necessary, this method will connect to the registry to get the
+// current digest of the tagged image. This function should not be used
+// for creating attestations in general.
```
klautcomputing: Do you want to explain here why it shouldn't be used? Binauthz operates on digests not tags :)

cmd/voucher_client/main.go
```diff
@@ -0,0 +1,48 @@
+package main
```
klautcomputing: Why don't we use cobra for this?

cmd/voucher_client/output.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/Shopify/voucher"
+)
+
+// formatResponse returns the response as a string.
+func formatResponse(resp *voucher.Response) string {
```
klautcomputing: I think for sake of consistency with other parts of voucher using logrus here would be a good idea :)

client/signimage.go
catherinejones: That's a good point! I will bump it to a couple of minutes.

cmd/voucher_client/main.go
catherinejones: Mostly for simplicity, since the client wasn't going to have any extra commands and is otherwise very small. Because Cobra uses the pflag library for flag support, so I figured it would make this command feel like the other two commands without adding the extra complexity. If you'd prefer I can definitely switch it to use cobra!

cmd/voucher_client/attest.go
catherinejones: Mostly so we could format the output more easily (Logrus adds a lot of extra information the output.

For example, calling `voucher_client` currently returns output similar to this:

```
 - Attesting image: imageregistry.com/path/to/image@sha256:de8c6942b6c909bb1adeaea8b7f16df04aca195b6b1b573feb9c1f3814d5792
   ✓ diy succeeded
   ✗ nobody failed
   ✓ snakeoil succeeded, but wasn't attested, err: rpc error: code = AlreadyExists desc = Requested entity already exists
```

cmd/voucher_client/digest.go
catherinejones: OK!

cmd/voucher_client/attest.go
catherinejones: I can merge the other pull request into this one if you are happy with the other! :)

klautcomputing: Let's merge the other one first and then update this one :)

PR #8, opened 2018-08-29
refactors authentication code

catherinejones:
This change refactors authentication against the container registry to no longer require the `gcloud` binary. This change also refactors the authorization code to allow for the creation of non-google authentication options in the future.


Dockerfile
```diff
@@ -4,20 +4,22 @@ FROM golang:1.10-alpine as builder
...
-&& apk add make \
-&& make all
+RUN apk --no-cache add \
+    git \
+    make && \
+    make all
 
 # Final build
-FROM google/cloud-sdk:alpine
+FROM alpine:3.7
```
klautcomputing: ❤️

Dockerfile
```diff
@@ -4,20 +4,22 @@ FROM golang:1.10-alpine as builder
 LABEL maintainer "klautcomputing"
 WORKDIR /go/src/github.com/Shopify/voucher
 COPY . .
-RUN apk update \
-&& apk add git \
-&& apk add make \
-&& make all
+RUN apk --no-cache add \
```
klautcomputing: 👍

docker/uri.go
```diff
@@ -5,11 +5,13 @@ import (
 	"net/url"
 
 	"github.com/docker/distribution/reference"
+	digest "github.com/opencontainers/go-digest"
 )
 
 // GetTokenURI gets the token URI for the passed repository.
 func GetTokenURI(ref reference.Named) string {
-	hostname, repository := reference.SplitHostname(ref)
+	hostname := reference.Domain(ref)
```
klautcomputing: would it make sense to put this out into a function?

catherinejones: I want to refactor all of the URI functions, so that they have the same base function that uses the "net/url" and "path" libraries instead of "fmt.Sprintf", but I haven't gotten to doing that.

authcheck.go
```diff
@@ -0,0 +1,9 @@
+package voucher
+
+// AuthCheck represents a Voucher check that needs to be authorized.
+// For example, a check that needs to connect to the registry will
+// need to implement AuthCheck.
+type AuthCheck interface {
```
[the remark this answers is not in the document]
catherinejones: Good point! I've renamed it to "AuthorizedCheck".

clair/scanner.go
```diff
@@ -23,17 +25,17 @@ func (scanner *Scanner) FailOn(severity voucher.Severity) {
 func (scanner *Scanner) Scan(i voucher.ImageData) ([]voucher.Vulnerability, error) {
 	vulns := make([]voucher.Vulnerability, 0)
 
-	gcloudToken, err := voucher.GetAccessToken()
-	if nil != err {
-		return vulns, err
-	}
+	// We set a longer timeout for this, given that this operation is far more
+	// intensive.
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
```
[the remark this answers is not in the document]
klautcomputing: Can we add some logging here in case of a timeout, so that we can notice it happening?

docker/uri.go
catherinejones: Sounds good to me!

clair/scanner.go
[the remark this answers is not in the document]
catherinejones: Any http.Clients using context.WithTimeout() will return an error if the timeout is missed! So an error will be returned to the client:

```json
{"name":"snakeoil","error":"Get https://imageregistry.com/v2/path/to/image/manifests/sha256:04bf3fbcf7e732e3a76f12a560fc04ad08f523d7061525755601b99d357a394e: context deadline exceeded","success":false,"attested":false}
```

Would this be acceptable, or should we process the error before returning it?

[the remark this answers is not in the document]
klautcomputing: As long as we can see it somewhere in our logging infrastructure that's okay 👍

docker/uri.go
klautcomputing: https://github.com/Shopify/voucher/issues/11

PR #15, opened 2018-09-11
adds support for encrypting clair login in ejson

catherinejones:
This change adds support for connecting to clair with BasicAuth.


config/secrets.production.ejson
```diff
@@ -5,5 +5,9 @@
         "nobody": "the pgp private key goes here",
         "provenance": "the pgp private key goes here",
         "snakeoil": "the pgp private key goes here",
+    },
+    "clair": {
+        "username": "username",
+        "password": "password here"
```
owencraston: will we need to fill this in?

catherinejones: Opps, it looks like my response wasn't posted... This is just an example, but it would need to be populated in the production copy of the secrets.ejson file.

PR #29, opened 2018-10-17
adds goreleaser support

catherinejones:
This change adds support for goreleaser, which can be called with `goreleaser` or `make release`.


Makefile
```diff
@@ -1,25 +1,22 @@
 # Go parameters
 GOCMD=go
 GODEP=dep
+GORELEASER=goreleaser
```
klautcomputing: Do we want to have a `make deps` that checks/installs things for you? Checks for go and docker and installs/updates goreleaser and dep?

cmd/voucher_cli/main.go
```diff
@@ -1,5 +1,12 @@
 package main
 
+var (
+	version = "dev"
+	commit  = "none"
+	date    = "unknown"
+)
+
 func main() {
+	rootCmd.Version = version
```
klautcomputing: What are we doing with this set version? Does cobra have an inbuilt version command?

.gitignore
```diff
@@ -12,10 +12,7 @@
 *.out
 
 # voucher specific output
-voucher_client
```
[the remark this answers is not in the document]
catherinejones: They were also ignoring the contents of `cmd/voucher_{cli,client,server}` and I couldn’t find a simpler way of blocking the binaries without moving them to a different directory.

Makefile
catherinejones: Just using `go get`? And then use the `$GOPATH` to find the binaries?

cmd/voucher_cli/main.go
catherinejones: It does! If you set the version it automatically enables it!

.gitignore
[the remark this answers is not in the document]
klautcomputing: Let's make a new dir then! 🙂 `/exe(cutables)`?

Makefile
klautcomputing: I thought something like:
```
DOCKER := $(shell command -v dot 2> /dev/null)

deps:
ifndef DOCKER
error "Docker not installed"
endif
```
and for the other ones either install or update. How does that sound to you?

.gitignore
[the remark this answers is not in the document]
catherinejones: What about `/build`? Since we already have `/dist` as a result of this change!

Makefile
catherinejones: We could definitely do that! I can add that to this pull request?

.gitignore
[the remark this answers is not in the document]
catherinejones: Added!

Makefile
catherinejones: I've added `make system-deps` for checking installed dependencies (since we have `make update-deps` for updating with `go dep`)!

PR #36, opened 2019-02-28
Add Clair Testing framework

anthonyho007:
Changes:
- Implemented Clair mock API server for testing
- Refactored some clair functions to ease testing
- Added unit tests to test our clair api


clair/scanner.go
```diff
@@ -72,3 +64,30 @@ func NewScanner(config Config, auth voucher.Auth) *Scanner {
 
 	return scanner
 }
+
+func getDockerManifest(ctx context.Context, tokenSrc oauth2.TokenSource, image reference.Canonical) (schema2.Manifest, error) {
+	client := oauth2.NewClient(ctx, tokenSrc)
+	return docker.RequestManifest(client, image)
+}
+
+func (scanner *Scanner) getVulnerabilities(manifest schema2.Manifest, tokenSrc oauth2.TokenSource, i voucher.ImageData) ([]voucher.Vulnerability, error) {
```
catherinejones: Could I suggest we move the `getClairVulnerabilities` method back into the `Scan` method, so we can remove some of the extra parameters? This function could be renamed to `convertVulnerabilities()` and take the passed Clair vulnerabilities and convert them to Voucher vulnerabilities?

testing/clair.go
```diff
@@ -0,0 +1,83 @@
...
+	user, pass, _ := req.BasicAuth()
+	if user != basicUsername || pass != basicPassword {
+		http.Error(writer, fmt.Sprintf("Unauthorized: %s", req.URL.Path), 401)
+	}
+	switch req.URL.Path {
+	case "/v1/layers":
+		var data map[string]v1.Layer
+		_ = json.NewDecoder(req.Body).Decode(&data)
+		mock.digestLayers[data["Layer"].Name] = true
+		fmt.Print(data["Layer"].Name)
```
catherinejones: Could remove the fmt.Print* lines here?

testing/clair.go
```diff
@@ -0,0 +1,83 @@
...
+						{
+							Name:        "bad vul 1",
+							Description: "some bad vul that i cant comprepend",
+							FixedBy:     "Hackerman",
+							Severity:    "Medium",
+						},
+						{
+							Name:        "bad vul 2",
+							Description: "Dont know what this is",
+							FixedBy:     "Hackerman",
```
catherinejones:
```suggestion
							FixedBy:     "Hackerwoman",
```

So that we don't have only one vulnerability fixer! 😜

clair/clair.go
```diff
@@ -98,24 +97,16 @@ func getLayerFromClair(config Config, digest digest.Digest) (layer v1.Layer, err
 	return
 }
 
-// getVulnerabilities gets a map[string]v1.Vulnerability from Clair, so that we can convert
+// getClairVulnerabilities gets a map[string]v1.Vulnerability from Clair, so that we can convert
 // them to Voucher Vulnerabilities all at once.
-func getVulnerabilities(ctx context.Context, config Config, tokenSrc oauth2.TokenSource, image reference.Canonical) (map[string]v1.Vulnerability, error) {
+func getClairVulnerabilities(manifest schema2.Manifest, config Config, tokenSrc oauth2.TokenSource, image reference.Canonical) (map[string]v1.Vulnerability, error) {
```
catherinejones: Great idea! This really simplifies this function!

clair/clair_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	clairScanner := newClairScanner(clairServer)
+	tokenSrc, _ := clairScanner.auth.GetTokenSource(context.TODO(), img)
+
+	vulnerabilities, err := clairScanner.getVulnerabilities(vtesting.NewTestManifest(), tokenSrc, img)
+
+	assert.Nil(t, err)
+	assert.Equal(t, 2, len(vulnerabilities))
+
+	for _, vulnerability := range vulnerabilities {
```
catherinejones: Rather than doing this, could I suggest that we use a table of `Vulnerability` structs and run:

```golang
require.Equal(t, expectedVulnerabilities, vulnerabilities)
```

This will save us from looping through the vulnerabilities!

testing/clair.go
```diff
@@ -0,0 +1,83 @@
...
+		_ = json.NewDecoder(req.Body).Decode(&data)
+		mock.digestLayers[data["Layer"].Name] = true
+		fmt.Print(data["Layer"].Name)
+
+		respond(writer, jsonContentType, "okay")
+		return
+	default:
+		for key := range mock.digestLayers {
+			if "/v1/layers/"+key == req.URL.Path {
+				respond(writer, jsonContentType, newClairLayerWithVulnerabilities())
```
catherinejones: Would it be possible to return different vulnerabilities for each layer?

testing/clair.go
```diff
@@ -0,0 +1,83 @@
...
+// NewTestClairServer creates a mock of Clair
+func NewTestClairServer(t *testing.T) *httptest.Server {
+	handler := new(clairAPIMock)
+	handler.digestLayers = map[string]bool{}
+
+	server := httptest.NewServer(handler)
+	return server
+}
+
+func newClairLayerWithVulnerabilities() v1.LayerEnvelope {
```
catherinejones: Would it be possible to pass this function from the Clair tests?

For example, in `clair/clair_test.go`:

```golang
img := vtesting.NewTestReference(t)
clairServer := vtesting.NewTestClairServer(t, newClairLayerFunc)
defer clairServer.Close()
```

Then in `testing/clair.go:

```golang
	default:
		for key := range mock.digestLayers {
			if "/v1/layers/"+key == req.URL.Path {
				respond(writer, jsonContentType, mock.layerFunc()) 
```

What do you think?

clair/clair_test.go
anthonyho007: Done

testing/clair.go
anthonyho007: Done

clair/scanner.go
```diff
@@ -72,3 +64,30 @@ func NewScanner(config Config, auth voucher.Auth) *Scanner {
...
+	vulns := make([]voucher.Vulnerability, 0)
+	rawVulns, err := getClairVulnerabilities(manifest, scanner.config, tokenSrc, i)
+	if nil != err {
+		return vulns, err
+	}
+
+	vulns = make([]voucher.Vulnerability, 0, len(rawVulns))
+	for _, rawVuln := range rawVulns {
+		if "" == rawVuln.Name {
+			fmt.Println("Empty Vulnerability???")
```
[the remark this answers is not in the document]
anthonyho007: Done

clair/scanner.go
anthonyho007: oh yeah i like ur idea! I have made the changes.

testing/clair.go
anthonyho007: I also added Hackerdog and Hackercat haha

testing/clair.go
anthonyho007: Yes!

testing/clair.go
anthonyho007: i think instead of passing in the function, i would pass in []v1.Vulnerability

testing/vulnerability.go
```diff
@@ -0,0 +1,80 @@
+package vtesting
+
+import (
+	"github.com/Shopify/voucher"
+	v1 "github.com/coreos/clair/api/v1"
+)
+
+// ClairVulnerabilitiesV1 return a list of clair vulnerabilities
+func ClairVulnerabilitiesV1() []v1.Vulnerability {
```
catherinejones: Could we combine these two functions into one function?

For example:

```golang
func ClairVulns(layer string) []v1.Vulnerability {
    var vulnsByLayer := map[string][]v1.Vulnerability{
        "abc1234": []v1.Vulnerability{
            {
               Name: "bad vuln",
            },
            {
               Name: "bad vuln 2",
            },
        },
        "anotherhash": []v1.Vulnerability{
            {
               Name: "bad vuln 3",
            },
            {
               Name: "bad vuln 4",
            },
        },
    }

    return vulnsByLayer[layer]
}
```

This would save us from having more than one function, and I'd argue it might help with understanding what's going on.

clair/scanner.go
```diff
@@ -35,23 +38,18 @@ func (scanner *Scanner) Scan(i voucher.ImageData) ([]voucher.Vulnerability, erro
...
-			vulns = append(vulns, vuln)
-		}
+	clairVulns, err := getClairVulnerabilities(manifest, scanner.config, tokenSrc, i)
+	if nil != err {
+		return vulns, err
 	}
 
+	vulns = convertToVoucherVulnerabilities(clairVulns, scanner.failOn)
+
 	return vulns, err
```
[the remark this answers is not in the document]
anthonyho007: Done

testing/vulnerability.go
```diff
@@ -0,0 +1,80 @@
+package vtesting
```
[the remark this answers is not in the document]
anthonyho007: true i will move them

testing/vulnerability.go
anthonyho007: ohh true and this is what u mean by table

clair/vulnerabilities_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"github.com/Shopify/voucher"
+	v1 "github.com/coreos/clair/api/v1"
+)
+
+// ClairVulnerabilitiesV1 return a list of clair vulnerabilities
+func ClairVulnerabilities(key string) []v1.Vulnerability {
+	vulns := map[string][]v1.Vulnerability{
+		"v1": []v1.Vulnerability{
```
catherinejones: Could you update all occurrences of `v1` to be one of the hashes in the manifest created by `vtesting.NewTestManifest()`

For example, `sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f` for `v1` and `sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736` for `v2`. As each represent a separate layer, are simulating four vulnerabilities, two added in each of the two layers (none added in the remaining layer).

clair/clair_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func TestGetClairVulnerabilites(t *testing.T) {
+	img, tokenSrc, clairServer := PrepareClairTest(t, ClairVulnerabilities("v1"))
+	defer clairServer.Close()
+
+	clairVulns, err := getClairVulnerabilities(vtesting.NewTestManifest(), createClairConfig(clairServer.URL), tokenSrc, img)
+	voucherVulns := convertToVoucherVulnerabilities(clairVulns, voucher.MediumSeverity)
+
+	assert.Nil(t, err)
+	assert.Equal(t, 2, len(voucherVulns))
```
catherinejones: In lines with my comment below, `len(voucherVulns)` should total 4 vulnerabilities, as we have two new issues in each layer.

testing/clair.go
```diff
@@ -0,0 +1,82 @@
...
+				mock.vulns = mock.vulns[1:]
+			}
+
+		}
+
+		respond(writer, jsonContentType, "okay")
+		return
+	default:
+		for key := range mock.layerVulns {
+			if "/v1/layers/"+key == req.URL.Path {
```
catherinejones: I'm afraid I missed this earlier, but this should return a different number of vulnerabilities depending on the layer returned. For example, if you request the last layer (see [this code](https://github.com/Shopify/voucher/blob/master/clair/clair.go#L130) for details), we should get all of the vulnerabilities not removed by subsequent layers.

clair/vulnerabilities_test.go
anthonyho007: Done

clair/clair_test.go
anthonyho007: it will be 3 since we filter out the lower vuln

testing/clair.go
anthonyho007: Yeah, i like this way better too!

PR #37, opened 2019-04-08
Limit IdleConnTimeout to 10s

Nilay-Shah:
This is done by introducing a new UpdateIdleConnectionsTimeout function
which will introduce this functionality.


auth/connections.go
```diff
@@ -0,0 +1,50 @@
...
+	}).DialContext,
+	MaxIdleConns:          100,
+	IdleConnTimeout:       10 * time.Second,
+	TLSHandshakeTimeout:   10 * time.Second,
+	ExpectContinueTimeout: 1 * time.Second,
+}
+
+// ErrCannotChangeTransport is an error returned when the Transport
+// is unable to be updated.
+var ErrCannotChangeTransport = errors.New("cannot update transport")
```
catherinejones: In retrospect I think this should be updated to match the function that calls it, something along the lines of "ErrCannotUpdateIdleConnTimeout"?

auth/connections.go
```diff
@@ -0,0 +1,50 @@
...
+
+// UpdateIdleConnectionsTimeout limits the default timeout for idle connections
+// to 10 seconds. If the OAuth2 transport is nil, redefine it with our
+// own DefaultTransport.
+func UpdateIdleConnectionsTimeout(client *http.Client) error {
```
catherinejones: I think we should probably give a bit more context here, for what it's fixing:

"This should limit the number of idle connections left open after the initial requests are made, which in turn should reduce the chances of the number of available connections being depleted."

auth/connections.go
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+// DefaultTransport is a custom implementation of the
+// DefaultTransport from the net/http stdlib. It limits
+// the IdleConnTimeout to 10 seconds instead of 90 seconds.
+var DefaultTransport http.RoundTripper = &http.Transport{
```
catherinejones: Let's throw the file that we copied the DefaultTransport from here!

```suggestion
//
// from "net/http/transport.go"
var DefaultTransport http.RoundTripper = &http.Transport{
```

PR #38, opened 2019-04-15
adds release instructions

catherinejones:
This change adds instructions for making a Voucher release!


CONTRIBUTING.md
```diff
@@ -63,3 +63,59 @@ are adding code which would be untested, please consider adding tests to cover
...
+
+First, update the master branch to the commit that you'd like to create a
+release for.
+
+```shell
+$ git checkout master
+$ git pull
+```
+
+If everything looks good, you will want to create a new tag. Voucher uses
```
klautcomputing:
```suggestion
If everything looks good, create a new tag. Voucher uses
```

CONTRIBUTING.md
```diff
@@ -63,3 +63,59 @@ are adding code which would be untested, please consider adding tests to cover
...
+digit. For example, if we are fixing a bug that doesn't affect other systems
+we can bump from v1.0.0 to v1.0.1. If the change is backwards compatible with
+the previous version, we can bump from v1.0.0 to v1.1.0. If the change is
+not backwards compatible, we must bump the version from v1.0.0 to v2.0.0.
+
+Run the following, where `version` is replaced with the appropriate version for
+this release:
+
+```shell
+$ git tag -s <version>
```
klautcomputing: What happens if someone just tags without signing?

CONTRIBUTING.md
```diff
@@ -63,3 +63,59 @@ are adding code which would be untested, please consider adding tests to cover
...
+```
+
+(Note that you will need to have Git configured to sign tags with 
+your OpenPGP key or this command will fail.)
+
+Before pushing your tag, you will want to build release versions of Voucher.
```
klautcomputing:
```suggestion
Before pushing your tag, build release versions of Voucher.
```

CONTRIBUTING.md
```diff
@@ -63,3 +63,59 @@ are adding code which would be untested, please consider adding tests to cover
...
+If this step fails, you will want to fix it, and that would require deleting
```
klautcomputing:
```suggestion
If this step fails, please fix it, and that would require deleting
```

CONTRIBUTING.md
catherinejones: I don't think there's a requirement that we sign tags but I'd probably argue we should so that we can ensure that nobody tampers with them after the fact!

klautcomputing: Totally agree! Just asked because I was curious whether github now has a feature where tags need to be signed.

CONTRIBUTING.md
```diff
@@ -63,3 +63,62 @@ are adding code which would be untested, please consider adding tests to cover
...
+If everything looks good, create a new tag. Voucher uses
+[Semantic Versioning](https://semver.org) which basically means that API
+compatible changes should bump the last digit, backwards compatible changes
+should bump the second, and API incompatible changes should bump the first
+digit. For example, if we are fixing a bug that doesn't affect other systems
+we can bump from v1.0.0 to v1.0.1. If the change is backwards compatible with
+the previous version, we can bump from v1.0.0 to v1.1.0. If the change is
+not backwards compatible, we must bump the version from v1.0.0 to v2.0.0.
+
+Run the following, where `version` is replaced with the appropriate version for
```
Nilay-Shah: I think it makes sense to put all of this in a numbered list? That way it's easy to tell when a step starts and ends. It also allows for less text to read because it would be:

1. `command_to_run`
description of `command_to_run` and some gotcha's that one may encounter

CONTRIBUTING.md
```diff
@@ -63,3 +63,62 @@ are adding code which would be untested, please consider adding tests to cover
...
+this release:
+
+```shell
+$ git tag -s <version>
+```
+
+(Note that you will need to have Git configured to sign tags with 
+your OpenPGP key or this command will fail.)
```
Nilay-Shah: Which command are you referring to? I think this will be solved if we number the steps, though

CONTRIBUTING.md
```diff
@@ -63,3 +63,62 @@ are adding code which would be untested, please consider adding tests to cover
...
+
+Before pushing your tag, build a release version of Voucher to ensure that
+everything builds properly:
+
+```shell
+$ make release
+```
+
+If this step fails, please do not make a release without fixing it and delete
```
Nilay-Shah: "...please do not make a release without fixing it. Delete the tag so it can..." reads better IMO

CONTRIBUTING.md
```diff
@@ -63,3 +63,62 @@ are adding code which would be untested, please consider adding tests to cover
...
+Next, push the tag to the server, where `version` is the same version you
+specified before:
+
+```shell
+$ git push origin refs/tags/<version>
+```
+
+Finally, create a new release in GitHub for the new version for the tag you
+created and signed, upload the binary tar archives that were created in `dist/`
+as well as the `changelog.txt` file. The output of the `CHANGELOG.md` file
```
Nilay-Shah: What actually changes in this `changelog.txt` file? This is the first time it's mentioned, so I'm not sure if the file is generated automatically or I have to generate it myself?

CONTRIBUTING.md
catherinejones: This is referring to:

```shell
$ git tag -s <version>
```

CONTRIBUTING.md
catherinejones: Do you have an example of documentation that does this? I'm not opposed to doing this, but I'm uncertain what it would look like.

CONTRIBUTING.md
catherinejones: I think this is a good indication that regardless of what we do this is not clearly linked to that command!

CONTRIBUTING.md
Nilay-Shah: Not off the top of my head, but I got my inspiration from https://github.com/Shopify/k8s-template-validator#releasing-a-new-version-shopify-employees since it's short and sweet

CONTRIBUTING.md
catherinejones: It's automatically generated by the goreleaser command!

catherinejones: I've reworded this, how does the reworded version sound?

Nilay-Shah: "...upload the automatically created..." to emphasize it was automatically created? Also your updated versions refers to `checksum.txt` and not `changelog.txt`; is that intentional?

CONTRIBUTING.md
catherinejones: I think this might be something we disagree on to be honest, I find that documentation too short and lacking in context as to why certain things are happening 😅

klautcomputing: I think more words don't hurt. Also if you look at the file rendered you basically have that, just execute only the commands and skip the text.

CONTRIBUTING.md
catherinejones: Yep! I had the wrong filename in the original version. I've made a couple more changes, do you think this is clear enough?

catherinejones:
Sorry for invalidating your review @klautcomputing 😅

PR #39, opened 2019-04-26
Handles images that have both tag and digest present

klautcomputing:
In case you have an image that has both tag and digest present Voucher will create an attestation for: gcr:tag@digest, this the attestation will actually be created in grafeas but is useless because kritis is looking for an attestation for gcr@digest. This PR fixes it by making sure Voucher slices out the digest in the resrouce uri.


grafeas/client.go
```diff
@@ -106,17 +110,26 @@ func (g *Client) AddAttestationToImage(reference reference.Canonical, payload vo
...
+		return nil, fmt.Errorf("Could not make %s a Named", ref)
+	}
+	digested, ok := ref.(reference.Digested)
+	if !ok {
+		return nil, fmt.Errorf("Could not make %s a Digested", ref)
+	}
+
 	resource := grafeaspb.Resource{
-		Uri: "https://" + reference.String(),
+		Uri: "https://" + named.Name() + "@" + digested.Digest().String(),
```
catherinejones: If we have a `reference.Canonical` (which we should) you should be able to do this without the conversion step!

grafeas/client.go
```diff
@@ -106,17 +110,26 @@ func (g *Client) AddAttestationToImage(reference reference.Canonical, payload vo
 	return item, err
 }
 
-func (g *Client) getCreateOccurrenceRequest(reference reference.Reference, parentNoteID string, attestation *grafeaspb.Occurrence_Attestation) *grafeaspb.CreateOccurrenceRequest {
+func (g *Client) getCreateOccurrenceRequest(ref reference.Reference, parentNoteID string, attestation *grafeaspb.Occurrence_Attestation) (*grafeaspb.CreateOccurrenceRequest, error) {
```
catherinejones: I believe this could be updated to use `reference.Canonical` as we’re using those everywhere else!

PR #41, opened 2019-05-01
handles non-json responses

catherinejones:
If Voucher returns non-JSON encoded data, we should not try to parse it with the JSON decoder.

This converts non-JSON Voucher responses to errors.


catherinejones:
> Is `Contains` necessary? Should we be stricter about the content-type?

This will match if we include character encoding extensions in the Content-Type, which I don't think we currently return but maybe we should? I'm not opposed to simplifying in this way!

> ReadAll is maybe more expressive?

I initially considered using ioutil.ReadAll but I figured it'd be essentially the same under the hood. Looking under the hood, it has some useful error handling so I think this change makes sense!

> Do yoda errors make sense? Assignment doesn't return a value in Golang
  
I didn't realize this but I'll probably keep writing them this way because I occasionally write C/C++!

> I usually try to return nil if there's an error because otherwise calling code can just throw away
> errors to "work around" them
  
Good point! I generally assume that people handle errors correctly but that's not a great assumption!

PR #44, opened 2019-05-10
fixes issue preventing us from loading image configs from GCR

catherinejones:
This change fixes a regression in v1.2.1, which prevented some image configs from loading from GCR.

This issue stems from our attempt to prevent non-JSON encoded data from being interpreted by the JSON decoder. Unfortunately, GCR returns `application/ocetet-stream` for certain image configurations, which breaks this logic.

This changes that logic to check the header for `json` or `application/ocetet-stream`. Short of attempting to guess the type of data based on the first few characters this strikes me as a good balance.


docker/call.go
```diff
@@ -5,25 +5,42 @@ import (
...
+func doDockerCall(client *http.Client, request *http.Request, data interface{}) error {
 	resp, err := client.Do(request)
 	if nil != err {
 		return err
 	}
 
 	defer resp.Body.Close()
 
-	if resp.StatusCode < 300 && resp.Header.Get("Content-Type") == contentType {
+	if resp.StatusCode < 300 && isJSONContent(resp.Header.Get("Content-Type")) {
```
catherinejones: I am open to changing this to include the `resp.Header.Get("Content-Type") == contentType` logic as well, but I think this is a good starting compromise.

docker/call.go
```diff
@@ -5,25 +5,42 @@ import (
...
 		return json.NewDecoder(resp.Body).Decode(&data)
 	}
 
 	return responseToError(resp)
 }
 
+// isJSONContent is used to prevent non-JSON data from being parsed by the json
+// decoder.
+func isJSONContent(contentType string) bool {
```
catherinejones: Alternatively, we could make this:

```golang
func isValidContentType(contentType, expectedContentType string) bool
```

Which might make more sense!

klautcomputing: Given that you check not only for "json" but also for "application/octet-stream" that might be the better name.

PR #45, opened 2019-07-06
Repository source feature

ItsMeWithTheFace:
This PR adds the capability of connecting a repository source (for now GitHub, but there are abstractions to potentially add other sources like GitLab or BitBucket) to obtain various information on the source code of a container image.

As a result, it can be integrated with other checks (which will be added in future PRs)


repository/github/client.go
```diff
@@ -0,0 +1,25 @@
...
+	"golang.org/x/oauth2"
+)
+
+// ClientGitHub represents the GitHub implementation of repository.RepoClient
+type ClientGitHub struct {
+	client *githubv4.Client
+}
+
+// NewGitHubClient creates a new GitHub client
+func NewGitHubClient(token string) repository.RepoClient {
```
catherinejones: Is there a reason we wouldn't pass a Context to NewGitHubClient? Since our contexts are usually set with a timeout, this might result in checks failing before a Github calls finishes, potentially leaving that connection open.

repository/github/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"github.com/djoksimo/githubv4"
+	"golang.org/x/oauth2"
+)
+
+// ClientGitHub represents the GitHub implementation of repository.RepoClient
+type ClientGitHub struct {
```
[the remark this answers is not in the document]
ItsMeWithTheFace: Yeah definitely!

repository/objects.go
```diff
@@ -0,0 +1,23 @@
+package repository
+
+// Organization contains repository information pertaining to an organization
+type Organization struct {
+	id   string
```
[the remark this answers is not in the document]
ItsMeWithTheFace: Done and done

repository/github/client.go
ItsMeWithTheFace: That's true, it makes more sense that a higher level module makes/passes in a context to the GitHub client at runtime too

repository/github/queries.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/Shopify/voucher/repository"
+	"github.com/djoksimo/githubv4"
+)
+
+// GetOrganization retrieves the necessary GitHub organizational information used in Voucher's checks
+func (ghc *ClientGitHub) GetOrganization(uri string) (*repository.Organization, error) {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
ItsMeWithTheFace: Wouldn't that mean we have to store the context as part of the Client o:? In the GoDocs for [`context`](https://godoc.org/context), they recommend against it since it should be used in functions that make external server calls. What do you think?

Nilay-Shah:
This may have been answered/discussed earlier, but why are we calling it `roundTripperWrapper` vs. `roundTripWrapper`?

catherinejones:
I think it's because it makes it more clear what it's wrapping (`RoundTripper`). `roundTripWrapper` sounds better but it's less clear that it's linked to `RoundTripper`.

repository/github/consts.go
```diff
@@ -0,0 +1,89 @@
...
+	organizationType = "Organization"
+	// commitType is one of GitHub's GraphQL schema types representing a Git commit
+	commitType = "Commit"
+)
+
+// CheckConclusionState is a string that represents the state for a check suite or check run conclusion.
+// CheckConclusionState is a type in the GitHub v4 GraphQL Schema
+type CheckConclusionState string
+
+func (s CheckConclusionState) isValidCheckConclusionState() bool {
```
djoksimo: I'd like to type the query objects with their respective types instead of `string` in `repository/objects.go`, should I export these as they're in different packages?

`repository/objects.go`:
```go
// CommitInfo contains the relevant information applied on a commit
type CommitInfo struct {
	URI         string
	CheckSuites []CheckSuite
	StatusState string
	IsSigned    bool
}

// CheckSuite is a collection of the check runs created by a single CI/CD App for a specific commit
type CheckSuite struct {
	App        CheckAppName
	Status     string
	Conclusion string
}
```

catherinejones: I'd suggest making generalized versions of those in the `repository` package, and export those, and make internal versions for the `github` package and convert to the `repository` versions!

catherinejones:
Are we getting panics locally? From the Travis CI:

```
?   	github.com/Shopify/voucher/repository	[no test files]
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7b46e3]
goroutine 7 [running]:
testing.tRunner.func1(0xc42011c2d0)
	/home/travis/.gimme/versions/go1.10.8.linux.amd64/src/testing/testing.go:742 +0x567
panic(0x809e60, 0xa4ddd0)
	/home/travis/.gimme/versions/go1.10.8.linux.amd64/src/runtime/panic.go:502 +0x24a
github.com/Shopify/voucher/repository/github.TestGetCommitInfo.func1(0xc42011c2d0)
	/home/travis/gopath/src/github.com/Shopify/voucher/repository/github/client_test.go:72 +0x1b3
testing.tRunner(0xc42011c2d0, 0xc42002cc60)
	/home/travis/.gimme/versions/go1.10.8.linux.amd64/src/testing/testing.go:777 +0x16e
created by testing.(*T).Run
	/home/travis/.gimme/versions/go1.10.8.linux.amd64/src/testing/testing.go:824 +0x565
```

I think this might be an indication of a bug in the library as much as an issue with the test suite...

ItsMeWithTheFace:
> Are we getting panics locally?
Panics aren't happening locally. The issues are most likely appearing because the `GITHUB_API_TOKEN` env variable isn't set. We still need to setup the authorization for tests in Travis.

catherinejones:
I don't believe non-existent environment variables should result in a panic though, so I think we need some logic to catch the environment variable not being set or whatever variable that is ending up nil before we use it?

repository/github/client_test.go
```diff
@@ -0,0 +1,128 @@
...
+			expectedAssociatedPullRequest: &repository.PullRequest{
+				BaseBranchName: "master",
+				HeadBranchName: "readme-updates",
+				IsMerged:       true,
+			},
+		},
+	}
+	for _, test := range getCommitInfoTests {
+		t.Run(test.testName, func(t *testing.T) {
+			token := os.Getenv("GITHUB_ACCESS_TOKEN")
```
catherinejones: I'd argue we should be reading the token in from the flags, so that we don't have to rely on an environment variable being set (and documented).

ItsMeWithTheFace:
Closing this PR since its been split into 2 different feature PRs

PR #46, opened 2019-08-03
Add repository client

ItsMeWithTheFace:
This PR adds the capability of connecting a repository source (for now GitHub, but there are abstractions to potentially add other sources like GitLab or BitBucket) to obtain various information on the source code of a container image.

As a result, it can be integrated with other checks (which will be added in future PRs)


repository/objects.go
```diff
@@ -0,0 +1,104 @@
...
+
+// CICDCheck is a collection of the check runs created by a single CI/CD App for a specific commit
+type CICDCheck struct {
+	App        CICDApp
+	Status     string
+	Conclusion string
+}
+
+// CICDApp contains the relevant information associated with a CI/CD app
+type CICDApp struct {
```
catherinejones: What does having "CICD" in front of Check and App tell us that `repository.Check` and `repository.App` not?

repository/objects.go
```diff
@@ -0,0 +1,104 @@
+package repository
+
+// Organization contains repository information pertaining to an organization
+type Organization struct {
+	ID   string
```
catherinejones: I think golint will complain about these fields not having comments explaining them!

catherinejones:
Could you also take a look at why the tests are failing on this PR?

repository/auth.go
```diff
@@ -0,0 +1,29 @@
+package repository
+
+// AuthKeyRing contains all the authentication keys
+// needed to communicate an org's repository source
+type AuthKeyRing map[string]Auth
```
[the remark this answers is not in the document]
ItsMeWithTheFace: Makes sense!

repository/objects.go
ItsMeWithTheFace: I could add in documentation for these fields, but golint hasn't complained about this o: Even with a `min_confidence` flag set to `0.0`. Maybe we should add linting to Travis as well for the future?

catherinejones: I think linting in Travis-CI makes a lot of sense! And if golint isn't complaining I won't!

repository/auth.go
```diff
@@ -0,0 +1,26 @@
...
+func (ac *AuthConfig) Type() string {
+	if ac.Token != "" {
+		return TokenAuthType
+	}
+
+	if ac.Username != "" && ac.Password != "" {
+		return UserPasswordAuthType
+	}
+
+	return ""
```
Nilay-Shah: Should we consider returning `nil` here?

repository/README.md
```diff
@@ -0,0 +1,5 @@
+# Source Code Repositories
+
+Voucher is capable of retrieving commit metadata from different version control sources (i.e., GitHub, GitLab, GitTea, etc.).
+
+At the moment, Voucher supports GitHub metadata as a source.
```
Nilay-Shah: I'm confused - I expected to see GitHub support but I don't see the `RepoClient` interface being implemented anywhere. Am I missing something?

ItsMeWithTheFace: The GitHub support is in PR #47. The `RepoClient` will be used in some checks that will get pushed in a completely separate PR (very soon) (:

Nilay-Shah: Yeah.. I just saw that PR! Makes sense. I thought it would be a part of this PR since it was still in the README, but since you'd be merging both very soon that should be fine!

repository/auth.go
ItsMeWithTheFace: I think we can only return nil if we're returning a `*string`, which could add some more complexity if we changed it o:

ItsMeWithTheFace:
CodeCov finally cooperating <:

rxbchen:
Nice :D!!!

PR #47, opened 2019-08-03
Add GitHub client

ItsMeWithTheFace:
This PR enables a user to connect a GitHub repository source to integrate with future checks


repository/github/queries.go
```diff
@@ -0,0 +1,282 @@
...
+func createNewDefaultBranch(defaultBranchName string, allDefaultBranchCommits []commit) *repository.DefaultBranch {
+	commits := make([]repository.Commit, 0)
+	for _, commit := range allDefaultBranchCommits {
+		commits = append(commits, *repository.CreateNewCommit(commit.URL))
+	}
+	return repository.CreateNewDefaultBranch(defaultBranchName, commits)
+}
+
+// commitCheckQuery is the GraphQL query for retrieving GitHub CI/CD status info for a specific commit
+type commitInfoQuery struct {
```
catherinejones: Could we move these structures into their own file? So `commit_info_query.go` and `default_branch_query.go`?

repository/github/queries.go
```diff
@@ -0,0 +1,282 @@
...
+	defaultBranchName := queryResult.Resource.Commit.Repository.DefaultBranchRef.Name
+
+	return createNewDefaultBranch(defaultBranchName, allDefaultBranchCommits), nil
+}
+
+// createNewDefaultBranch returns a pointer to a populated repository.DefaultBranch object
+func createNewDefaultBranch(defaultBranchName string, allDefaultBranchCommits []commit) *repository.DefaultBranch {
+	commits := make([]repository.Commit, 0)
+	for _, commit := range allDefaultBranchCommits {
+		commits = append(commits, *repository.CreateNewCommit(commit.URL))
```
catherinejones: Why are we doing `*repository.CreateNewCommit(commit.URL))` and not `repository.CreateNewCommit(commit.URL))`? Is `repository.CreateNewCommit(commit.URL))` returning the right type? Is `commits` the right type?

repository/github/queries.go
```diff
@@ -0,0 +1,282 @@
...
+	queryResult := new(defaultBranchQuery)
+	return getDefaultBranchFromQuery(ctx, ghc, queryResult, defaultBranchInfoVariables)
+}
+
+// getDefaultBranchFromQuery is the GraphQL query for retrieving GitHub default branch information
+func getDefaultBranchFromQuery(ctx context.Context, ghc *clientGitHub, queryResult *defaultBranchQuery, defaultBranchInfoVariables map[string]interface{}) (*repository.DefaultBranch, error) {
+	// GitHub has a limit of 100 records so we must perform pagination
+	var allDefaultBranchCommits []commit
+
+	for {
```
catherinejones: Could we abstract the pagination logic, rather than have it in this function?

repository/github/consts.go
```diff
@@ -0,0 +1,103 @@
...
+	repositoryType = "Repository"
+	// organizationType is one of GitHub's GraphQL schema types representing a GitHub organization
+	organizationType = "Organization"
+	// commitType is one of GitHub's GraphQL schema types representing a Git commit
+	commitType = "Commit"
+)
+
+// CheckConclusionState is a string that represents the state for a check suite or check run conclusion.
+// CheckConclusionState is a type in the GitHub v4 GraphQL Schema
+type CheckConclusionState string
```
catherinejones: Why are we exporting this? Do we need to use it elsewhere?

repository/github/consts.go
```diff
@@ -0,0 +1,103 @@
...
+	const (
+		completedString  = "COMPLETED"
+		inProgressString = "IN_PROGRESS"
+		queuedString     = "QUEUED"
+		requestedString  = "REQUESTED"
+	)
+	switch s {
+	case "": // CheckStatusState can be inconclusive/empty
+		return true
+	case completedString:
```
catherinejones: Could we not do:

```suggestion
	case completedString, inProgressString, queuedString, requestedString:
```

repository/github/client.go
```diff
@@ -0,0 +1,71 @@
...
+
+// GetOrganization retrieves the necessary GitHub organizational information used in Voucher's checks
+func (ghc *clientGitHub) GetOrganization(ctx context.Context, uri string) (*repository.Organization, error) {
+	repoInfo, err := newRepositoryOrgInfoResult(ctx, ghc, uri)
+	if err != nil {
+		return nil, err
+	}
+
+	if repoInfo.Resource.Typename != commitType {
+		return nil, fmt.Errorf(
```
catherinejones: This is a minor nitpick but here we're using a formatted error message to wrap the output of our custom error type from the `repository` package. We're losing some information by doing that. It'd be better if we could wrap the `TypeMismatchError` with a different error instead.

Gopkg.toml
```diff
@@ -61,3 +61,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "golang.org/x/crypto"
+
+[[constraint]]
+  branch = "master"
```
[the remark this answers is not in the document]
djoksimo: There seem to be no tagged releases for this package and the README states: 

> Status: In research and development. The API will change when opportunities for improvement are discovered; it is not yet frozen.

repository/github/consts.go
djoksimo: I didn't know that was possible in Go 🤔

Gopkg.toml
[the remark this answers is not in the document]
catherinejones: That's a little concerning honestly... Let's tag it against the latest commit hash on master for the time being.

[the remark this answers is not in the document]
djoksimo: >If you're interested in using the GraphQL API v4, the recommended library is shurcooL/githubv4.

https://github.com/google/go-github

repository/github/client.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+	"fmt"
+
+	"github.com/Shopify/voucher/repository"
+	"github.com/shurcooL/githubv4"
+	"golang.org/x/oauth2"
+)
+
+// ClientGitHub represents the GitHub implementation of repository.RepoClient
+type clientGitHub struct {
```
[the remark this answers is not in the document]
djoksimo: I think we chose to name this clientGithub to avoid struct name stuttering `client.client`, but I prefer your suggestion.

repository/github/queries.go
djoksimo: `repository.CreateNewCommit(commit.URL)` returns a pointer to a `repository.Commit` object so I'm dereferencing it. The other `repository.CreateNew<x>()` functions in the repository package return pointers as well.

 The slice of commits is of type `[]repository.Commit`:

```go
commits := make([]repository.Commit, 0)
```

Do you have any suggestions on changing/improving this?

repository/github/client.go
ItsMeWithTheFace: We could make a custom error type for this. I added the formatting to make it easier to determine where the error came from, but a stack trace would still be able to point that out as well.

With that being said, I could just return the typeMismatchError itself 😮

repository/github/queries.go
```diff
@@ -0,0 +1,208 @@
...
+	"context"
+	"fmt"
+
+	"github.com/Shopify/voucher/repository"
+	"github.com/shurcooL/githubv4"
+)
+
+// paginationQuery paginates through GitHub records and populates a destination slice with the appropriately typed query results
+// GitHub has a limit of 100 records so we must perform pagination
+func paginationQuery(ctx context.Context, ghc *client, dest interface{}, cursorName string, queryResult interface{}, queryPopulationVariables map[string]interface{}) (interface{}, error) {
```
djoksimo: I abstracted the pagination logic here and I am aware that it's messy so I'd appreciate your feedback on this @ItsMeWithTheFace @catherinejones  😄

repository/github/queries.go
```diff
@@ -0,0 +1,205 @@
...
+	for {
+		err := ghc.ghClient.Query(ctx, queryResult, queryPopulationVariables)
+		if err != nil {
+			return nil, err
+		}
+		var hasMoreResults bool
+
+		switch destination := dest.(type) {
+		case []checkSuite:
```
catherinejones: Why are we worrying about the types of the results here? I'd imagine this would be much simpler if you looped through these and returned them to a function rather than having the conversion in this function!

For example:

```
type QueryHandler func(queryResult *queryResultType) (bool, error)
```

And then here you could have:

```golang
	for {
		err := ghc.ghClient.Query(ctx, queryResult, queryPopulationVariables)
		if err != nil {
			return nil, err
		}
		var hasMoreResults bool

		hasMoreResults, err = queryHandler(queryResult)
 		if nil != err {
			return err
		}
		
		if !hasMoreResults {
			return nil
		}
	}
```

Then you could have your calling function have this logic:

```golang
v, err := paginationQuery(ctx, ghc, , cursorName string, queryPopulationVariables, func(v interface{}) (bool, error) {
                        commit := queryResult.(*commitInfoQuery).Resource.Commit
			resourceType := queryResult.(*commitInfoQuery).Resource.Typename
			if resourceType != commitType {
				return false, repository.NewTypeMismatchError(commitType, resourceType)
			}
			dest = append(destination, commit.CheckSuites.Nodes...)
			hasMoreResults = commit.CheckSuites.PageInfo.HasNextPage
			queryPopulationVariables[cursorName] = githubv4.NewString(commit.CheckSuites.PageInfo.EndCursor)
                return hasMoreResults, err
})

File: repository/github/queries.go (hunk 2 above)
Comment (catherinejones) [reply repository/github/queries.go]: I think you might be explaining the "what" when I was asking for the "why". I understand what's going on in the code, I'm wondering if we should be using `repository.Client` in two different ways. If `repository.CreateNew<x>()` returns a pointer, why wouldn't we be treating it as a pointer everywhere?

File: repository/github/queries.go (hunk 10 above)
Comment (djoksimo) [reply repository/github/queries.go]: Good suggestion! 😸

File: repository/github/queries_test.go [hunk 11]
```diff
@@ -0,0 +1,796 @@
...
+			input:     new(repositoryOrgInfoQuery),
+			mask:      []string{},
+			expected:  nil,
+		},
+	}
+
+	for _, test := range newRepoOrgTests {
+		t.Run(test.testName, func(t *testing.T) {
+			c := new(mockGitHubGraphQLClient)
+			c.HandlerFunc = func(query interface{}, variables map[string]interface{}) error {
```
catherinejones: Could we maybe split this out into a function?

For example:

```golang
func createHandler(commitURL string, input *repositoryOrgInfoQuery, mask []string, expected  *repositoryOrgInfoQuery) HandlerFunc {
    // the code in question
}
```

Alternatively we could make the `struct` that contains the test data into a type and have that be the argument to `createHandler`? What do you think?

repository/github/queries.go
```diff
@@ -0,0 +1,225 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Shopify/voucher/repository"
+	"github.com/shurcooL/githubv4"
+)
+
+const queryPageLimit = 3
```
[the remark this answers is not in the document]
ItsMeWithTheFace: Good idea!

repository/github/default_branch_query.go
```diff
@@ -0,0 +1,30 @@
+package github
+
+import "github.com/shurcooL/githubv4"
+
+// defaultBranchQuery is the GraphQL query for retrieving information pertaining to the repository's default branch
+type defaultBranchQuery struct {
```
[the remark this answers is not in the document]
ItsMeWithTheFace: In my editor it's using an indentation size of 2, but I think it'd be nice if we have a `.editorconfig` file at the root to keep everything consistent between different editors. What do you think?

repository/github/consts.go
```diff
@@ -0,0 +1,82 @@
...
+	organizationType = "Organization"
+	// commitType is one of GitHub's GraphQL schema types representing a Git commit
+	commitType = "Commit"
+)
+
+// checkConclusionState is a string that represents the state for a check suite or check run conclusion.
+// checkConclusionState is a type in the GitHub v4 GraphQL Schema
+type checkConclusionState string
+
+func (s checkConclusionState) isValidCheckConclusionState() bool {
```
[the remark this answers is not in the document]
ItsMeWithTheFace: True, I think I might add it to the existing `utils.go` file in that case

repository/github/default_branch_query.go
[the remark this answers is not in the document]
Nilay-Shah: Yeah that would be cool!

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
```
Nilay-Shah: Wouldn't this set it for the whole project and not the `*_query.go` files?

repository/github/queries.go
```diff
@@ -0,0 +1,225 @@
...
+
+// queryHandler is called on every iteration of paginationQuery to populate a slice of query results
+// queryHandler checks to see whether there are more records given that GitHub has a limit of 100 records per query
+type queryHandler func(queryResult interface{}) (bool, error)
+
+// paginationQuery populates a destination slice with the appropriately typed query results
+// GitHub has a limit of 100 records so we must perform pagination
+func paginationQuery(ctx context.Context, ghc ghGraphQLClient, queryResult interface{}, queryPopulationVariables map[string]interface{}, pageLimit int, qh queryHandler) error {
```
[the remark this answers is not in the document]
djoksimo: Sure 👍

.editorconfig
ItsMeWithTheFace: It would, but I'm not sure if that's what we want. I could make it for just the `*_query.go` files though

PR #48, opened 2019-08-12
Pull Build Metadata from Grafeas

rxbchen:
This PR added the functionality of pulling build metadata from Grafeas. The metadata can be used later to generate/validate attestations.


build_detail.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"errors"
+	"net/url"
+	"path"
+	"regexp"
+)
+
+const (
+	githubRegex = "^(?:https?|git)(?:://|@)(?P<vcs>[^/:]+)[/:](?P<org>[^/.]+)/(?P<name>[^/.]+)(?:.git)?"
```
catherinejones: I think this regex should be moved into `repository/github`!

build_detail.go
```diff
@@ -0,0 +1,70 @@
...
+
+// RepositoryMetadata describes the top level metadata information about a repo
+// that one can get from the gitUrl
+type RepositoryMetadata struct {
+	Vcs          string `json:"vcs"`
+	Organization string `json:"organization"`
+	Name         string `json:"name"`
+}
+
+func parseGithubURL(gitURL string) (*RepositoryMetadata, error) {
```
catherinejones: Please move this function into repository/github as well!

catherinejones:
👋 Hi @rxbchen! I hope you don't mind, @lynnsh is going to take over this PR and finish it up!

build_detail.go
rxbchen: Yup! Will do

rxbchen:
👋 Hey @catherinejones sound good! Let me know if you need any context/info from me @lynnsh

grafeas/client.go
```diff
@@ -58,8 +58,7 @@ func (g *Client) GetMetadata(ctx context.Context, reference reference.Canonical,
 			}
 			break
 		}
-		item := new(Item)
-		item.Occurrence = occ
+		item := OccurrenceToBuildDetails(occ)
```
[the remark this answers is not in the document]
lynnsh: Done.

grafeas/client.go
```diff
@@ -114,6 +114,33 @@ func (g *Client) Close() {
 	g.grafeas.Close()
 }
 
+// GetBuildDetails gets BuildDetails for the passed image.
+func (g *Client) GetBuildDetails(ctx context.Context, reference reference.Canonical) (items []voucher.BuildDetail, err error) {
+	filterStr := resourceURL(reference)
```
catherinejones: I think this will give us all metadata associated with `reference`, could we change this to 
```suggestion
	filterStr := filterStr = kindFilterStr(reference, common.NoteKind_BUILD)
```

This will only return the BuildDetails associated with `reference`!

build_detail.go
```diff
@@ -0,0 +1,57 @@
...
+// related to a build
+type BuildDetail struct {
+	RepositoryURL string `json:"repository"`
+	Commit        string `json:"commit"`
+	Author        string `json:"author"`
+	BuildURL      string `json:"build_url"`
+}
+
+// GetCommitURL generates a commit url from the build metadata
+func (b *BuildDetail) GetCommitURL() (string, error) {
```
catherinejones: I'm not certain this function belongs in this package, could we move it to `repository/github`?

I think it should be something like:

```golang
func GetCommitURL(buildDetails *voucher.BuildDetails) (string, error) {
```

What do you think?

build_detail_test.go
```diff
@@ -0,0 +1,69 @@
+package voucher
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/Shopify/voucher/repository/github"
+)
+
+func TestParseGithutURL(t *testing.T) {
```
catherinejones: Inline with my other comment, I think this is too github specific for the main Voucher package!

grafeas/client.go
lynnsh: Done

build_detail.go
lynnsh: Done

build_detail_test.go
lynnsh: Done (file renamed and moved to repository/github)

grafeas/build.go
```diff
@@ -0,0 +1,16 @@
+package grafeas
+
+import (
+	"github.com/Shopify/voucher"
+	grafeaspb "google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1/grafeas"
+)
+
+// OccurrenceToBuildDetails converts an Occurrence to a Build_Detail
+func OccurrenceToBuildDetails(occ *grafeaspb.Occurrence) (detail *voucher.BuildDetail) {
```
catherinejones: Since we're using pass by value in `GetBuildDetails`, why don't we return just `voucher.BuildDetail` here, rather than `*voucher.BuildDetail`?

lynnsh: Done.

checks/provenance/check.go
```diff
@@ -37,54 +32,37 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
...
-	// unmarshal against provenance bytes
-	if err := json.Unmarshal(provenance, details); err != nil {
-		log.Errorf("Provenance bytes do not match details: %v", err)
-		return
-	}
-
-	if !trustedBuilderIdentities[details.Provenance.Creator] {
-		log.Errorf("Builder identity not trusted: %s", details.Provenance.Creator)
+	if !trustedBuilderIdentities[detail.BuildCreator] {
+		log.Errorf("Builder identity not trusted: %s", detail.BuildCreator)
```
catherinejones: Could we change this function so it returns both the `trusted` boolean and these log lines as errors? I'd prefer if `check` didn't depend on logrus!

grafeas/build.go
```diff
@@ -0,0 +1,30 @@
...
+
+	detail.ProjectID = buildProvenance.GetProjectId()
+    detail.BuildCreator = buildProvenance.GetCreator()
+	detail.BuildURL = buildProvenance.GetLogsUri()
+	detail.RepositoryURL = buildProvenance.GetSourceProvenance().GetContext().GetGit().GetUrl()
+	detail.Commit = buildProvenance.GetSourceProvenance().GetContext().GetGit().GetRevisionId()
+
+	buildArtifacts := buildProvenance.GetBuiltArtifacts()
+
+    detail.Artifacts = make([]voucher.BuildArtifact, 0, len(buildArtifacts))
```
catherinejones: Could you run `go fmt` in the voucher source tree?

checks/provenance/check.go
```diff
@@ -37,54 +32,37 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
...
 		return true, nil
 	}
 
 	return false, nil
 }
 
-func validateProvenance(details *build.Details) (trusted bool) {
+func validateProvenance(detail voucher.BuildDetail) (trusted bool) {
 	// get trusted things
 	trustedBuilderIdentities := voucher.ToMapStringBool(viper.GetStringMap("trusted-builder-identities"))
```
catherinejones: Could we make a separate function for setting these? as well, if you could take a look at `cmd/config` and see how we pass arguments into the Checks that will help us figure out how we should pass these arguments without depending on Viper

checks/provenance/check.go
lynnsh: Done.

grafeas/build.go
lynnsh: Done.

checks/provenance/check.go
lynnsh: Done.

build_detail.go
```diff
@@ -0,0 +1,41 @@
...
+		str += "ProjectID: " + b.ProjectID + "\n"
+	}
+	str_artifacts := ""
+	for _, val := range b.Artifacts {
+		if val.String() != "" {
+			str_artifacts += val.String() + ","
+		}
+	}
+	if str_artifacts != "" {
+		str += "Artifacts: " + str_artifacts[:len(str_artifacts)] + "\n"
```
catherinejones: This is essentially the same as:

```suggestion
		str += "Artifacts: " + str_artifacts + "\n"
```

cmd/config/checks.go
```diff
@@ -59,6 +59,22 @@ func setCheckValidRepos(check voucher.Check, validRepos []string) {
...
+// trustedBuilderIdentities is a map of BuildCreators to bool, which is set to true for trusted build creators
+func setCheckTrustedBuilderIdentities(check voucher.Check, trustedBuilderIdentities map[string]bool) {
+	if provenanceCheck, ok := check.(voucher.ProvenanceCheck); ok {
+		provenanceCheck.SetTrustedBuilderIdentities(trustedBuilderIdentities)
+	}
+}
+
+// setCheckTrustedBuilderProjects sets trusted builder projects used in ProvenanceCheck;
+// trustedBuilderProjects is a map of projects to bool, which is set to true for trusted projects
+func setCheckTrustedBuilderProjects(check voucher.Check, trustedBuilderProjects map[string]bool) {
```
catherinejones: Could you combine these two functions into one?

catherinejones: We will also need to call this function on line 81, and pass it the values that we get from viper!

checks/provenance/check.go
```diff
@@ -37,54 +42,34 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
...
-
-	// unmarshal against provenance bytes
-	if err := json.Unmarshal(provenance, details); err != nil {
-		log.Errorf("Provenance bytes do not match details: %v", err)
-		return
-	}
-
-	if !trustedBuilderIdentities[details.Provenance.Creator] {
-		log.Errorf("Builder identity not trusted: %s", details.Provenance.Creator)
+func validateProvenance(p *check, detail voucher.BuildDetail) (trusted bool, logerr string) {
```
catherinejones: Rather than returning a string here we should return a Error, the way you are doing on line 49:

```suggestion
func validateProvenance(p *check, detail voucher.BuildDetail) (trusted bool, err error) {
```

checks/provenance/check.go
```diff
@@ -37,54 +42,34 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
...
+	if !p.trustedBuilderIdentities[detail.BuildCreator] {
+		logerr = fmt.Sprintf("Builder identity not trusted: %s", detail.BuildCreator)
```
catherinejones:
```suggestion
		err = Errorf("builder identity not trusted: %s", detail.BuildCreator)
```

build_detail.go
lynnsh: This is my error, it should be `str_artifacts[:len(str_artifacts)-1]` because I collect artifacts as strings separated by comma, and I wanted to remove the last dangling comma. Fixed now.

cmd/config/checks.go
lynnsh: Done.

checks/provenance/check.go
lynnsh: Done.

checks/provenance/check.go
lynnsh: Done.

build_detail.go
```diff
@@ -0,0 +1,41 @@
...
+	if b.BuildURL != "" {
+		str += "BuildURL: " + b.BuildURL + "\n"
+	}
+	if b.ProjectID != "" {
+		str += "ProjectID: " + b.ProjectID + "\n"
+	}
+	str_artifacts := ""
+	for _, val := range b.Artifacts {
+		if val.String() != "" {
+			str_artifacts += val.String() + ","
```
catherinejones: Rather than doing this, I might suggest doing `strings.Join()` instead of string concatenation here [strings.Join](https://golang.org/pkg/strings/#Join)!

checks/provenance/check.go
```diff
@@ -37,54 +38,33 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
...
 
-	item, ok := items[0].(*grafeas.Item)
-	if !ok {
-		return false, fmt.Errorf("response from MetadataClient is not an grafeas.Item")
-	}
-
-	buildDetails := item.Occurrence.GetBuild()
-	if validateProvenance(buildDetails) && validateArtifacts(i, buildDetails) {
-		log.Infof("Validated image provenance and artifacts for: %s", i.String())
+	buildDetail := items[0]
```
catherinejones: I think this might be fragile, could we loop through all the BuildDetails and validate each one?

If there's only one the loop will be equivalent to just getting the first one, and if there is more than one and one of those is not validatable, we probably have to investigate anyways!

cmd/config/checks.go
```diff
@@ -59,6 +60,15 @@ func setCheckValidRepos(check voucher.Check, validRepos []string) {
 	}
 }
 
+// setCheckTrustedBuilderIdentities sets trusted builder identities and projects used in ProvenanceCheck
+// trustedBuilderIdentities is a map of BuildCreators to bool, which is set to true for trusted build creators
+// trustedBuilderProjects is a map of projects to bool, which is set to true for trusted projects
+func setCheckTrustedBuilderIdentities(check voucher.Check, trustedBuilderIdentities map[string]bool, trustedBuilderProjects map[string]bool) {
+	if provenanceCheck, ok := check.(voucher.ProvenanceCheck); ok {
+		provenanceCheck.SetTrustedBuilderIdentities(trustedBuilderIdentities, trustedBuilderProjects)
```
catherinejones: `SetTrustedBuilderIdentities` is no longer accurate I'd argue, could we change it back to two separate functions?

cmd/config/checks.go
```diff
@@ -59,6 +60,15 @@ func setCheckValidRepos(check voucher.Check, validRepos []string) {
 	}
 }
 
+// setCheckTrustedBuilderIdentities sets trusted builder identities and projects used in ProvenanceCheck
+// trustedBuilderIdentities is a map of BuildCreators to bool, which is set to true for trusted build creators
+// trustedBuilderProjects is a map of projects to bool, which is set to true for trusted projects
+func setCheckTrustedBuilderIdentities(check voucher.Check, trustedBuilderIdentities map[string]bool, trustedBuilderProjects map[string]bool) {
```
catherinejones: This is good! Could we rename it to:
```suggestion
func setCheckTrustedIdentitiesAndProjects(check voucher.Check, trustedBuilderIdentities []string, trustedBuilderProjects []string) {
```

Since we're not just trusting identities, we're also trusting projects here!

provenancecheck.go
```diff
@@ -0,0 +1,8 @@
+package voucher
+
+// ProvenanceCheck represents a Voucher check that sets
+// trusted projects and build creators
+type ProvenanceCheck interface {
+	Check
+	SetTrustedBuilderIdentities(map[string]bool, map[string]bool)
```
catherinejones: As I mentioned in my previous comment I do think it's worth having two separate functions for setting the trusted builders and identities. Could you re-split these out?

As well, the function names are very long, could we shorten them to:

```suggestion
	SetTrustedBuildCreators([]string)
	SetTrustedProjects([]string)
```
?

checks/provenance/check.go
```diff
@@ -2,32 +2,33 @@ package provenance
...
 }
 
 // SetMetadataClient sets the MetadataClient for this Check.
 func (p *check) SetMetadataClient(metadataClient voucher.MetadataClient) {
 	p.metadataClient = metadataClient
 }
 
+// SetTrustedBuilderIdentities sets trustedBuilderIdentities and trustedBuilderProjects for this Check.
+func (p *check) SetTrustedBuilderIdentities(trustedBuilderIdentities map[string]bool, trustedBuilderProjects map[string]bool) {
+	p.trustedBuilderIdentities = trustedBuilderIdentities
```
catherinejones: I'm going to be Very Nitpicky here, could we change the function to:

```golang
func (p *check) SetTrustedBuildCreators(buildCreators []string) {
 	for _, identity := range buildCreators {  
 	 	p.trustedBuildCreators[identity] = true
 	}
}
```

The advantage of doing this is that we don't have to force the use of the `map[string]bool` structure elsewhere, and we can be consistent with how the DIY check works as well (it also uses `[]string]`).

cmd/config/checks.go
```diff
@@ -68,6 +78,9 @@ func NewCheckSuite(metadataClient voucher.MetadataClient, names ...string) (*vou
 	scanner := newScanner(metadataClient, auth)
 	checksuite := voucher.NewSuite()
 
+	trustedBuilderIdentities := voucher.ToMapStringBool(viper.GetStringMap("trusted-builder-identities"))
```
catherinejones: As per one of my other change requests this should change to `GetStringSlice` I believe!
```suggestion
	trustedBuilderIdentities := viper.GetStringSlice("trusted-builder-identities")
```

checks/provenance/check.go
catherinejones: Opps sorry, lets get rid of the `if` statement above!

build_detail.go
lynnsh: Done.

checks/provenance/check.go
lynnsh: Done.

cmd/config/checks.go
lynnsh: Done (in `check.go`).

cmd/config/checks.go
lynnsh: Done.

provenancecheck.go
lynnsh: Done.

cmd/config/checks.go
lynnsh: Done.

build_detail.go
```diff
@@ -0,0 +1,45 @@
...
+	if b.BuildCreator != "" {
+		str += "BuildCreator: " + b.BuildCreator + "\n"
+	}
+	if b.BuildURL != "" {
+		str += "BuildURL: " + b.BuildURL + "\n"
+	}
+	if b.ProjectID != "" {
+		str += "ProjectID: " + b.ProjectID + "\n"
+	}
+	str_artifacts := ""
```
catherinejones: Alternatively, you could start this as a slice of strings!

```suggestion
	str_artifacts := make([]string, 0, len(b.Artifacts))
```

Currently we're expanding the `str_artifacts` every time we change it, whereas we could be allocating one big string at the end of the call when we use `strings.Join`.

Issue #49, opened 2019-08-13
Readiness check to prevent Voucher from being called when connections are saturated


catherinejones:
Voucher currently has an issue where it (infrequently) consumes all of the potential network sockets for the system it's running in (usually a Pod in Kubernetes). Our solution to date is to delete the failing Voucher Pod so that Kubernetes starts a new Pod to replace it. This is very inefficient, and it would be preferable to have Kubernetes not send traffic to a pod if its' connections are saturated.

To this end, we should investigate setting up a [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-readiness-probes) API call, which we can use to prevent Voucher being sent traffic unnecessarily.

This check should measure if it can open new sockets or confirm if it's using too many. If it's close to or at the limit, it should return an error code.

chrisdonaldson:
Why is it consuming all the network sockets?

catherinejones:
I believe it's handling too many connections at once, and they're taking too long to finish. We've attempted to remedy this with https://github.com/Shopify/voucher/pull/37, however we still occasionally have problems and have to restart the Voucher instance.

catherinejones:
Checking a pod that just failed led me to a total of 28431 open sockets! It might be better to find why we have so many open sockets rather than adding this to the health check!

catherinejones:
Based on a quick experiment, it appears that the number of sockets being opened is much smaller and they appear to be closing much more quickly. I'm going to close this issue for now, and we can circle back to it if we want to consider this in the future!

PR #50, opened 2019-08-30
fix Grafeas connection issues

catherinejones:
This may or may not help us avoid implementing a Health Check API call for #49.

This change updates the way we interact with the MetadataClient, and more specifically Grafeas, to allow us to set the `context` on the individual grafeas API calls (which are [gRPC](https://en.wikipedia.org/wiki/GRPC) under the hood). The other side effect of this is that we're being more idiomatic in our handling of `context`s in `grafeas.Client` because we are no longer keeping it as a part of the `grafeas.Client` structure.

From the golang documentation:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

The net result of this change is that we need to pass the context around the application more. To handle that, I've updated the Check and CheckSuite types to allow for this.

Because this is an API breaking change, I think we should bump the major number to reflect that.


rxbchen:
That's a nice change! I remember that the docs did mention that there may be issues of you store context as a variable instead of passing it in.

PR #51, opened 2019-09-09
adds Interface, adds contexts to Client

catherinejones:
This change adds an `Interface` type to the Voucher package, and adds contexts to the Voucher client calls, rather than using a `timeout` value.


interface.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+
+	"github.com/docker/distribution/reference"
+)
+
+// Interface represents an interface to the Voucher API. Typically Voucher API
+// clients would implement it.
+type Interface interface {
```
klautcomputing: Could this maybe be named a little more meaningful, please?

catherinejones: I was following the pattern established by [kubernetes.Interface](https://godoc.org/k8s.io/client-go/kubernetes#Interface), but I can definitely rename it. Originally I had it as `Checker`, do you think that would be preferable?

klautcomputing: TIL. Feel free to merge then :) `Checker` would also be okay.

PR #54, opened 2019-09-15
Remove vendor/

epk:
Following up on #52

Go will ignore the vendor folder when using modules unless you call it with the flag `-mod=vendor` or set `GOFLAGS=-mod=vendor`

cc @catherinejones @klautcomputing


catherinejones:
Hi @epk! Thank you for your PR! I'll take a look at this sometime today! 😄

epk:
> Hi @epk! Thank you for your PR! I'll take a look at this sometime today! 😄

To save you some time 

https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

>By default, go commands like go build ignore the vendor directory when in module mode.
The -mod=vendor flag (e.g., go build -mod=vendor) instructs the go commands to use the main module's top-level vendor directory to satisfy dependencies. The go commands in this mode therefore ignore the dependency descriptions in go.mod and assume that the vendor directory holds the correct copies of dependencies. Note that only the main module's top-level vendor directory is used; vendor directories in other locations are still ignored.
Some people will want to routinely opt-in to vendoring by setting a GOFLAGS=-mod=vendor environment variable.

.travis.yml
```diff
@@ -1,9 +1,17 @@
 language: go
 
+cache:
```
epk: Caching `$GOPATH/pkg/mod` for faster build/CI runs to offset time added by not vendoring.

Makefile
```diff
@@ -44,8 +46,14 @@ clean:
...
 	done
 
+ensure-deps:
+	$(GOCMD) mod tidy
+	$(GOCMD) mod download
+	$(GOCMD) mod verify
+
 update-deps:
-	$(GOCMD) mod vendor
+	$(GOCMD) get -u -t all
```
epk: `go get -u -t all` will update all the dependencies to their latest patch versions. (`-t` to update test dependencies)

CONTRIBUTING.md
```diff
@@ -44,9 +44,11 @@ $ git checkout -b <the name of your branch>
...
 up your formatting/check for issues.
 
-    - If you update the dependencies, you may have to run `make update-deps` to
- ensure that the dependencies are vendored. This should happen in a separate
-commit from those containing your source modifications.
+    - If you add a dependencies, run `go mod tidy` to prune any no-longer-needed
+    dependencies from `go.mod`.
+
+    - To update dependencies to use newer minor or patch releases when available:
+    run `make update-deps`.
```
catherinejones: I think this needs to be updated to be:

```suggestion
    run `make ensure-deps`.
```

epk:
> Hi @epk! I'm really sorry for not getting to this earlier, I had to rapidly change where I was focusing so I haven't been thinking about Voucher as much the past couple of days.
> 
> A couple of questions:
> 
> * why did we change `update-deps` to `ensure-deps`? We're still updating the dependencies, are we not?
> * why did we not add `go.sum` as well?
> * did you hand edit the source files or was that the result of a go command? `go fmt -s`?
> 
> This looks really good! I do think we probably want to add the go.sum file before merging, just to be safe. Could you add that to the PR?

No worries 🙂 

When we were vendoring with `dep` `update-deps` (`dep ensure`) made sure `/vendor` is in sync with `Gopkg.{toml, lock}`, It was not really updating the dependencies per se. For that if I recall correctly the command is `dep ensure -update`

Thats why I switched the makefile to have two tasks, one to ensure the dependencies in the Go modules cache are in sync with `go.{mod, sum}` and another to update the dependencies. I _think_ this should also resolve https://github.com/Shopify/voucher/pull/54#discussion_r326210404

`go.sum` exists in my branch and has not changed

I ran `goimports -w -l -local github.com/Shopify/voucher .` to format the repo.

CONTRIBUTING.md
catherinejones: Opps, your comment in https://github.com/Shopify/voucher/pull/54#issuecomment-533201129 explained why this is not necessary!

catherinejones:
Sorry @epk I should have read the changes more carefully! I think this is good to go!

catherinejones:
Your PR has been merged @epk! Thank you for contributing to Voucher! 🎉

PR #56, opened 2019-09-25
refactor GetVulnerabilities and add it to MetadataClient

lynnsh:
Added GetVulnerabilities method to MetadataClient and removed getVulnerabilitiesForImage from grafeas/scanner (since it is refactored in grafeas/client as GetVulnerabilities).
This also allows to remove the dependecy on GetMetadata in grafeas/scanner.
Added logic to choose package vulnerability over the default vulnerability if it is not specified.


grafeas/scanner.go
```diff
@@ -21,29 +20,6 @@ func (s *Scanner) FailOn(severity voucher.Severity) {
...
-		if voucher.ShouldIncludeVulnerability(vuln, s.failOn) {
-			vulns = append(vulns, vuln)
-		}
-	}
-
-	return vulns, nil
-}
 
 // Scan gets the vulnerabilities for an Image.
 func (s *Scanner) Scan(ctx context.Context, i voucher.ImageData) ([]voucher.Vulnerability, error) {
```
catherinejones: Would it be possible to move the `pollForDiscoveries` logic into `GetVulnerabilites` as well?

grafeas/client.go
```diff
@@ -110,6 +111,27 @@ func (g *Client) getCreateOccurrenceRequest(reference reference.Canonical, paren
 	return req
 }
 
+// GetVulnerabilities returns the detected vulnerabilities for the Image described by voucher.ImageData.
+func (g *Client) GetVulnerabilities(ctx context.Context, reference reference.Canonical) ([]voucher.Vulnerability, error) {
+	items, err := g.GetMetadata(ctx, reference, voucher.VulnerabilityType)
```
catherinejones: Could we use the same pattern we used for GetBuildDetails function here? If possible we should remove references to the `GetMetadata` function so that we can deprecate it!

grafeas/scanner.go
lynnsh: Done

grafeas/client.go
lynnsh: Done

grafeas/client.go
```diff
@@ -110,6 +110,33 @@ func (g *Client) getCreateOccurrenceRequest(reference reference.Canonical, paren
 	return req
 }
 
+// GetVulnerabilities returns the detected vulnerabilities for the Image described by voucher.ImageData.
+func (g *Client) GetVulnerabilities(ctx context.Context, reference reference.Canonical) (items []voucher.Vulnerability, err error) {
+	filterStr := kindFilterStr(reference, common.NoteKind_BUILD)
```
catherinejones: In this case we should be using a different NoteKind, since this is a Vulnerability, not a Build!

lynnsh: Done

grafeas/vulnerability.go
```diff
@@ -34,7 +34,14 @@ func getSeverity(severity string) voucher.Severity {
 func OccurrenceToVulnerability(occ *grafeaspb.Occurrence) (vul voucher.Vulnerability) {
 	vul.Name = strings.Replace(occ.GetNoteName(), vulProject, "", 1)
 	vulnDetails := occ.GetDetails().(*grafeaspb.Occurrence_Vulnerability).Vulnerability
-	vul.Severity = getSeverity(vulnerability.Severity_name[int32(vulnDetails.Severity)])
+
+	severity := getSeverity(vulnerability.Severity_name[int32(vulnDetails.Severity)])
+
+	if severity == voucher.UnknownSeverity {
+		vul.Severity = getSeverity(vulnDetails.GetPackageIssue()[0].GetSeverityName())
```
catherinejones: We should double check that we have at least one PacakgeIssue here! 

```golang
if len(packageIssues) >= 1 {
    // etc
}
```

main/test.go
```diff
@@ -0,0 +1,75 @@
+package main
```
catherinejones: Could you remove this from the source tree? You can copy it out of this package and run it separately:

```
$ mv main/test.go ~/voucher_vulnearbility_test.go
$ git add main
```

You can then still run this by calling:

```
$ go run voucher_vulnearbility_test.go <arguments>
```

grafeas/vulnerability.go
```diff
@@ -34,7 +34,14 @@ func getSeverity(severity string) voucher.Severity {
...
+	} else {
+		vul.Severity = severity
```
catherinejones: Rather than having this else, we can simply update `vul.Severity` on line 38 and override it on line 41 if necessary!

grafeas/vulnerability.go
lynnsh: Done.

main/test.go
lynnsh: Done.

grafeas/vulnerability.go
lynnsh: Done.

PR #58, opened 2019-10-07
fixes linting errors

catherinejones:
This PR fixes some linter errors!

You can see these errors by installing `gometalinter`:

1. `go get -u github.com/golangci/golangci-lint/cmd/golangci-lint`
2. In the Voucher directory, run the following
    - `golangci-lint run --disable-all --enable=ineffassign --enable=misspell --enable=gocyclo --enable=vet --enable=gofmt --enable=golint`


checks/provenance/check.go
```diff
@@ -44,7 +44,7 @@ func (p *check) Check(ctx context.Context, i voucher.ImageData) (bool, error) {
 	}
 
 	for _, buildDetail := range items {
-		ok := true
+		var ok bool
```
chrisdonaldson: This does change functionality but looking at how it's handled that should be fine.

catherinejones: Because the next line:

```golang
ok, err = validateProvenance(p, buildDetail)
```

writes to the OK variable in the same scope, setting it to any value isn't useful because it will immediately be overwritten!

PR #60, opened 2019-10-30
GitHub RepositoryClient updates

lynnsh:
- added GetBranch method that queries a specified branch using GitHub API
- refactored CommitInfo into Commit and Commit into CommitRef
- updated repository.Client to accept BuildData instead of Commit
   - this created an import cycle: repository.Client requires voucher.BuildDetail and voucher.RepositoryCheck requires repository.Client
        - as a solution, BuildDetail and BuildArtifact were moved to repository package
- updated branch and org queries to search on repository url rather than a commit url
- added RepositoryCheck
- removed ID from Organization


repository/objects.go
```diff
@@ -2,22 +2,20 @@ package repository
 
 // Organization contains repository information pertaining to an organization
```
catherinejones: Could we remove the `Create` in the `CreateNew<x>` functions? Generally functions that create new values are `New<x>` in Go!

repository/github/queries.go
```diff
@@ -54,38 +54,42 @@ func newRepositoryOrgInfoResult(ctx context.Context, ghc ghGraphQLClient, uri st
 
 	queryResult := new(repositoryOrgInfoQuery)
 	if err := ghc.Query(ctx, queryResult, repoInfoVariables); err != nil {
-		return nil, fmt.Errorf("RepositoryInfo query could not be completed. Error: %s", err)
+		return repository.Organization{}, fmt.Errorf("RepositoryInfo query could not be completed. Error: %s", err)
+	}
+	if queryResult.Resource.Repository.Owner.Typename != organizationType {
```
catherinejones: Is there an `OrganizationType` we could leverage from the Github package?

repository/github/branch_query.go
```diff
@@ -0,0 +1,28 @@
+package github
+
+import "github.com/shurcooL/githubv4"
+
+// branchQuery is the GraphQL query for retrieving information pertaining to the repository's default branch
```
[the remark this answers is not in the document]
lynnsh: But this query retrieves information for the branch specified [here](https://github.com/Shopify/voucher/pull/60/files/e36bf93f6d760daa37081fafe760f7e5e0bc82a6#diff-27b1b4092cf5c6f56247185e513c3457R25), not for a default branch.

repository/github/queries.go
lynnsh: In voucher's Github package? This one (`organizationType`) comes from `repository/github/consts.go`. Should we use something else?

catherinejones: No I was thinking more of in the 3rd party GitHub package! Since there might be something in there already!

repository/github/branch_query.go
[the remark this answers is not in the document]
catherinejones: Yeah! I think the current comment is out of date 😅

repository/github/queries.go
lynnsh: I couldn't find anything that would be useful here (checked google, github and godoc).

repository/objects.go
lynnsh: Done.

repository/github/branch_query.go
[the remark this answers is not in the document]
lynnsh: Oh, right, sorry, fixed now.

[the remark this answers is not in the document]
catherinejones: No worries! :)

PR #61, opened 2019-10-30
Testing RepositoryClient and MetadataClient

lynnsh:
- added RepositoryClient mockup that could be used in testing
- added RepositoryClient interface that implements repository.Client interface with the addition of AddRepository function (used in testing, so that it is possible to add different repositories when testing with repositoryclient mockup; currently it's 1-1 relationship between repositories and organizations in repositoryclient mockup)
- added MetadataClient mockup to use in testing


testing/repository.go
```diff
@@ -0,0 +1,13 @@
+package vtesting
+
+import (
+	"github.com/Shopify/voucher/repository"
+)
+
+// Repository contains repository information pertaining to a repository
+type Repository struct {
+	Org      repository.Organization
+	Mame     string
```
catherinejones: What is `Mame` here?

lynnsh: A typo, it is supposed to be `Name`, sorry. Fixed now.

PR #62, opened 2019-10-30
Add Organization check

lynnsh:
- added OrgCheck that gets build data from grafeas using MetadataClient and compares it to the data provided by GitHub


cmd/config/repo.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func newRepositoryClient(org repository.Organization) (repositoryClient repository.Client) {
+	if strings.Contains(org.URL, "github.com") {
+		keyring, err := getRepositoryKeyRing()
+		if nil != err {
+			log.Errorf("failed to get github token: %s", err)
+			return nil
+		}
+		token := keyring["github"]
```
catherinejones: Could we change the token to match the domain of the server?
```suggestion
		token := keyring["github.com/Shopify"]
```

Then we can select the token based on the whole domain?

```golang
for domain, token := range keyring {
    if strings.Contains(org.URL, domain) {
         if isGithubDomain(org.URL) {
              return gitHub.NewClient(ctx, &token)
         }

    } 
}
```

lynnsh: The statement [before](https://github.com/Shopify/voucher/pull/62/files/65d50f709c403e33f30fac1059b99229ebb015db#diff-d121569490d02cdb701b44da27084ac9R14): `if strings.Contains(org.URL, "github.com") {` checks that it is a Github domain before getting the keyring. Should it be rechecked by `isGithubDomain`? Or would it require some other checks?

catherinejones: We might have situations where we have multiple GitHub organizations, which could require different GitHub tokens, especially when it comes to orgs that have multiple GitHub environ
ments (eg. the Grafeas org in Github vs the Google org in Github).

lynnsh: But what would `isGithubDomain(org.URL)` check do that this check `if strings.Contains(org.URL, "github.com")` doesn't? 
It doesn't seem like `org.URL` depends on loop variables, so why is this check inside?

catherinejones: > But what would isGithubDomain(org.URL) check do that this check if strings.Contains(org.URL, "github.com")

I think it's totally reasonable to just do that!

> It doesn't seem like org.URL depends on loop variables, so why is this check inside?

it doesn't necessarily need to be! But it might be easier to iterate over the available tokens rather than iterate over the different organizations, you know?

lynnsh: Done.

cmd/config/repo.go
```diff
@@ -17,13 +17,16 @@ func newRepositoryClient(org repository.Organization) (repositoryClient reposito
 			log.Errorf("failed to get github token: %s", err)
 			return nil
 		}
-		token := keyring["github"]
-		repositoryClient, err = github.NewClient(context.Background(), &token)
-		if nil != err {
-			log.Errorf("failed to connect to github: %s", err)
-			return nil
+		for domain, token := range keyring {
+		    if strings.Contains(org.URL, domain) {
```
catherinejones: Could we make a dedicated function that parses this? 

I just realized we might end up with a situation where someone gives their repository bad metadata so that we leak a token to the wrong API endpoint, so we should

for example:

```
Repository: catherinesmaliciousserver.com/github.com/Shopify/voucher
```

Then the malicious server could just implement the part of the Github API required to accept tokens and we'd leak tokens to them unintentionally! TLS wouldn't protect us because we'd probably use that domain as the valid domain name

We could prevent this by parsing the URL with `*url.Parse()` and then pass the resulting `*url.URL` to the URL check:

```golang
func isURL(orgURL, repoURL *url.URL) bool {
    if orgURL.Hostname() != repoURL.Hostname() {
        return false
    }
    return strings.HasPrefix(repoURL.EscapedPath(), orgURL.EscapedPath())
}
```

cmd/config/repo.go
```diff
@@ -0,0 +1,32 @@
...
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/Shopify/voucher/repository"
+	"github.com/Shopify/voucher/repository/github"
+)
+
+func newRepositoryClient(org repository.Organization) (repositoryClient repository.Client) {
+	if strings.Contains(org.URL, "github.com") {
```
catherinejones:
```suggestion
	 orgURL, err := url.Parse(org.URL)
	 if nil != err {
	 	 log.Errorf("failed to get github token: %s", err)
 		 return nil
 	}
 	
	if strings.HasSuffix(orgURL.Hostname, "github.com") {
```

Inline with my suggestion below, we should probably parse the URL so we don't end up treating `catherinesmaliciousserver.com/github.com/Shopify/voucher` as a GitHub URL!

cmd/config/repo.go
lynnsh: Done.

cmd/config/repo.go
lynnsh: Done.

cmd/config/repo.go
```diff
@@ -0,0 +1,46 @@
...
+					return nil
+				}
+				return repositoryClient
+			}
+		}
+	}
+	return nil
+}
+
+func isOrgURL(orgURL, repoURL *url.URL) bool {
```
catherinejones: I'm worried this code isn't doing what it is intended on doing...

From my brief experiments, when I call:

```golang
orgURL, err := url.Parse("https://github.com/Shopify")
```

The `orgURL.Hostname()` function returns `github.com`.

However, with 

```golang
orgURL, err := url.Parse("github.com/Shopify")
```

The `orgURL.Hostname()` function returns `` (an empty string).

This means that we could get both an empty string for the orgURL or repoURL hostname. I think we need to do some normalization to make sure we don't end up doing this.

We could, for instance, do something where we render and reparse the URL with `https://` in front of the domain?

```golang
orgURL, err := url.Parse("https://"+orgURL.String())
```

cmd/config/get_orgs_config.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"github.com/Shopify/voucher/repository"
+	"github.com/spf13/viper"
+)
+
+func GetOrganizationsFromConfig() (orgs map[string]repository.Organization) {
+	orgs = make(map[string]repository.Organization)
+	repositories, ok := viper.Get("repositories").([]interface{})
+	if !ok {
+		return
```
catherinejones: We should add some logging here to inform the user that there are no repositories if there are none.

We could also simplify this function by replacing this line with the following:

```suggestion
		repositories = []interface{}{}
```

This will mean that the loop on line 14 will not execute, and we can then log if `orgs` is empty.

cmd/config/repo_test.go
```diff
@@ -0,0 +1,27 @@
...
+	assert.True(t, github.IsGithubRepoClient(client), "received testing client and not a github client")
+}
+
+func TestTestingRepo(t *testing.T) {
+	viper.Set("ejson.secrets", "../../testdata/test_repo.ejson")
+	organization := r.Organization{Name: "TestOrg", URL: "https://gitlab.com/TestOrg"}
+	client := newRepositoryClient(organization)
+
+	assert.False(t, github.IsGithubRepoClient(client), "received github client and not a testing client")
+}
```
catherinejones: Could we add some tests for `isOrgURL` here?

cmd/config/repo.go
```diff
@@ -11,14 +12,20 @@ import (
 )
 
 func newRepositoryClient(org repository.Organization) (repositoryClient repository.Client) {
-	if strings.Contains(org.URL, "github.com") {
+	orgURL, err := url.Parse(org.URL)
+	 if nil != err {
+		log.Errorf("failed to get github token: %s", err)
```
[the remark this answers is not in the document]
lynnsh: Done.

cmd/config/checks.go
```diff
@@ -82,6 +83,13 @@ func NewCheckSuite(metadataClient voucher.MetadataClient, names ...string) (*vou
 	trustedBuildCreators := viper.GetStringSlice("trusted-builder-identities")
 	trustedProjects := viper.GetStringSlice("trusted-projects")
 
+	orgs := GetOrganizationsFromConfig()
+	for name, organization := range orgs {
+		repositoryClient := newRepositoryClient(organization)
+		orgCheck := org.NewOrganizationCheckFactory(organization, repositoryClient)
+		voucher.RegisterCheckFactory("is_"+name, orgCheck)
```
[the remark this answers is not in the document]
lynnsh: Done.

cmd/config/checks.go
```diff
@@ -82,6 +83,13 @@ func NewCheckSuite(metadataClient voucher.MetadataClient, names ...string) (*vou
 	trustedBuildCreators := viper.GetStringSlice("trusted-builder-identities")
 	trustedProjects := viper.GetStringSlice("trusted-projects")
 
+	orgs := GetOrganizationsFromConfig()
+	for name, organization := range orgs {
+		repositoryClient := newRepositoryClient(organization)
```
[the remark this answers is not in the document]
lynnsh: I added a check for nil that would skip creating OrgCheck for that client.

cmd/config/checks.go
```diff
@@ -86,8 +87,11 @@ func NewCheckSuite(metadataClient voucher.MetadataClient, names ...string) (*vou
 	orgs := GetOrganizationsFromConfig()
 	for name, organization := range orgs {
 		repositoryClient := newRepositoryClient(organization)
+		if repositoryClient == nil {
+			continue
```
catherinejones: Could we log here using `log.Errorf` that we failed to setup the repositoryClient for the organization?

cmd/config/repo.go
```diff
@@ -38,6 +38,14 @@ func newRepositoryClient(org repository.Organization) (repositoryClient reposito
 	return nil
 }
 
+func parseURL(urlStr string) (*url.URL, error) {
+	urlParsed, err := url.Parse(urlStr)
+	if err == nil && urlParsed.Hostname() == "" {
```
catherinejones: Since parsing the URL is more time consuming than doing string comparisons, could we use `strings.HasPrefix` to check for `http://` or `https://` and parse once?

cmd/config/repo.go
```diff
@@ -38,6 +38,14 @@ func newRepositoryClient(org repository.Organization) (repositoryClient reposito
 	return nil
 }
 
+func parseURL(urlStr string) (*url.URL, error) {
```
catherinejones: could we rename this to `normalizeURL`?

checks/org/check.go
```diff
@@ -0,0 +1,58 @@
...
+	items, err := o.metadataClient.GetBuildDetails(ctx, i)
+	if err != nil {
+		if voucher.IsNoMetadataError(err) {
+			return false, ErrNoBuildData
+		}
+		return false, err
+	}
+
+	for _, buildDetail := range items {
+		org, err := o.repositoryClient.GetOrganization(ctx, buildDetail)
```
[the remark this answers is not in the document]
lynnsh: `repositoryClient` is created when this check is created that has nil check added. So at the moment this would not happen. Should there be an additional check in OrgCheck in case there is another way to create this check that would cause `repositoryClient` be nil?

cmd/config/get_orgs_config.go
lynnsh: Done

cmd/config/checks.go
lynnsh: Done.

cmd/config/repo.go
lynnsh: Done.

checks/org/check_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	repoClient, _ := vtesting.NewClient()
+	commits := map[string]r.Commit{
+		"cdef0987": {URL: "https://github.com/TestOrg/app/commit/cdef0987", Checks: nil, Status: "good", IsSigned: true,
+			AssociatedPullRequests: nil},
+	}
+	branches := map[string]r.Branch{
+		"master": {Name: "master", CommitRefs: []r.CommitRef{{URL: "https://github.com/TestOrg/app/commit/cdef0987"}}},
+	}
+	repoClient.AddRepository(r.CreateNewOrganization("TestOrg", "https://github.com/TestOrg"), "TestRepo", commits, branches)
```
catherinejones:
```suggestion
	repoClient.AddRepository(r.NewOrganization("TestOrg", "https://github.com/TestOrg"), "TestRepo", commits, branches)
```

checks/org/check_test.go
```diff
@@ -0,0 +1,81 @@
...
+			AssociatedPullRequests: nil},
+		"efgh6543": {URL: "https://github.com/Shopify/app/commit/efgh6543", Checks: nil, Status: "great", IsSigned: false,
+			AssociatedPullRequests: []r.PullRequest{{BaseBranchName: "base", HeadBranchName: "master", IsMerged: false},
+				{BaseBranchName: "branch-base", HeadBranchName: "test", IsMerged: true}}},
+	}
+	branches := map[string]r.Branch{
+		"test":   {Name: "test", CommitRefs: []r.CommitRef{{URL: "https://github.com/Shopify/app/commit/efgh6543"}}},
+		"master": {Name: "master", CommitRefs: []r.CommitRef{{URL: "https://github.com/Shopify/app/commit/cdef0987"}}},
+	}
+	repoClient.AddRepository(r.CreateNewOrganization("Shopify", "https://github.com/Shopify"), "Repo_Shopify", commits, branches)
```
catherinejones: I just realized this is in an issue since we changed `CreateNewOrganization` to `NewOrganization` in the other PR, could you update this to:

```suggestion
	repoClient.AddRepository(r.NewOrganization("Shopify", "https://github.com/Shopify"), "Repo_Shopify", commits, branches)
```

checks/org/check_test.go
catherinejones:
```suggestion
	repoClient.AddRepository(r.CreateNewOrganization("TestOrg", "https://github.com/TestOrg"), "TestRepo", commits, branches)
```

checks/org/check_test.go
catherinejones: Since we changed `CreateNewOrganization` to `NewOrganization` in the previous PRs, we must update these references:

```suggestion
	repoClient.AddRepository(r.NewOrganization("Shopify", "https://github.com/Shopify"), "Repo_Shopify", commits, branches)
```

checks/org/check_test.go
lynnsh: Changed to `r.CreateNewOrganization` to `r.NewOrganization`.

checks/org/check_test.go
lynnsh: Done.

Issue #59, opened 2019-10-21
MetadataClient for Open Source Grafeas


catherinejones:
At the time that we started the Voucher project Grafeas hadn’t had a general release that we could build off of, so we stuck to the Container Analysis API that Google made available to us. 

Now that [Grafeas v0.1.3](https://github.com/grafeas/grafeas/releases/tag/v0.1.3) is available, it’s worth having a MetadataClient that communicates with it directly instead. We can later merge the ContainerAnalysis (now confusingly named “grafeas”) and Grafeas MetadataClients together.

Since this will be a breaking change, we should rename the ContainerAnalysis MetadataClient to something else.

- Locally
    - [ ] Setup a local instance of Grafeas (feel free to create a docker image)
    - [ ] Populate it with test data!
    - [ ] Create a script (ideally in Go, so it can be used in the tests) for starting up Grafeas and populating it with data
- In Voucher
    - [ ] Create a new MetadataClient for Grafeas (“grafeasos” for “grafeas open source”?)
    - [ ] Add configuration options to select MetadataClient at startup
    - [ ] Add unit tests and integration tests (that tests against the local copy of Grafeas)

Nilay-Shah:
Work still left to do reusing #63:
- [ ] change Grafaes version back to `v1beta1`
- [ ] remove vendored code
- [ ] rebase and retest

Nilay-Shah:
This is blocked because of some incorrect functionality in the `v1beta` code.

Issue #21, opened 2018-09-27
Remove stringly typed check


klautcomputing:
`isVulnerabilityDiscovery` found here https://github.com/Shopify/voucher/pull/17/files#diff-000ec35b674e1e00d32ecbb98f0dd556R11 should not rely on a string comparison but actually check the type.

catherinejones:
This may not be possible with the state of permissions for the `ContainerAnalysis` API. I will confirm with Google during our sync today.

catherinejones:
As per our contact at Google, we are not able to access this note and should use the note path as the way to determine if it's a vulnerability note!

PR #63, opened 2019-11-15
MetadataClient grafeas open source

lynnsh:
For issues #59 and #21.
- added the script file `grafeas_test_setup.go` that sets up the local instance of grafeas (using docker) and populates it with test data (notes and occurrences)
- added MetadataClient for Grafeas open source (using [client-go](https://github.com/grafeas/client-go) library)
- added tests for the grafeas MetadataClient
- modified `grafeas[containeranalysis]/GetMetadata` to return occurrences of the same kind as their note’s kind
- modified `grafeas[containeranalysis]/isVulnerabilityDiscovery` to use noteKind for comparison instead of the noteName
- added the option to select grafeasos MetadataClient from the config file
- renamed `grafeas` to `containeranalysis`
- moved scanner from `containeranalysis` to `scanner` and renamed `Scanner` to `MetadataScanner`
- removed `GetMetadata()` function from MetadataClient interface and structs that implement it

Notes:
- added `vendor/` folder, so tests should be run including it (`go test ./... -mod=vendor`) since there are a few issues in client-go that need to be resolved, so for now they are resolved locally
    - the list of issues:
        - json formatting (underscores instead of camelCase that the server expects)
        - invalid server url formatting (https://github.com/grafeas/client-go/issues/13)
        - (doesn't require `vendor/` folder) filtering is not implemented on the server side
        - (doesn't require `vendor/` folder) weird [status code check](https://github.com/grafeas/client-go/blob/master/0.1.0/api_grafeas_v1_beta1.go#L97) for the client (`localVarHttpResponse.StatusCode == 200` is nested in `localVarHttpResponse.StatusCode >= 300`, so it would never be executed). It does not seem to cause any issues, but still it might be a bug


grafeasos/client.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	return payload, err
+}
+
+// GetMetadata gets metadata of the requested type for the passed image.
+func (g *Client) GetMetadata(ctx context.Context, reference reference.Canonical, metadataType voucher.MetadataType) (items []voucher.MetadataItem, err error) {
+	kind := getNoteKind(metadataType)
+	var occurrences []Item
+	if kind == grafeaspb.DISCOVERY_V1beta1NoteKind {
+		occurrences, err = g.getNotesOccurrencesForKind(ctx, kind, isDiscoveryVulnerabilityNote)
```
catherinejones: This might be too fragile a function currently... what if I tried to get a Discovery for Deployments or something else?

I think it might be worth removing this GetMetadata, given that we have other functions to replace it and it's going to cause these problems? What do you think?

grafeasos/client.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	return
+}
+
+// NewClient creates a new Grafeas Client.
+func NewClient(ctx context.Context, imageProject, binauthProject string, keyring *voucher.KeyRing) (*Client, error) {
+	client := new(Client)
+	config := grafeaspb.NewConfiguration()
+	config.BasePath = "http://localhost:8080"
```
catherinejones: I don't think we should be hard-coding the Grafeas path here... we should add this as an argument or pass in a struct containing configuration details for Grafeas!

scanner/scanner.go
```diff
@@ -1,27 +1,27 @@
...
 
 	"github.com/Shopify/voucher"
 )
 
-// Scanner implements voucher.VulnerabilityScanner, and connects to Grafeas
+// MetadataScanner implements voucher.VulnerabilityScanner, and connects to Grafeas
 // to obtain vulnerability information. It will block while scanning is active
 // and fail if it spends more than a minute waiting for discovery to finish.
-type Scanner struct {
+type MetadataScanner struct {
```
catherinejones: Would it be possible to move this into the Voucher package?

grafeasos/item.go
```diff
@@ -0,0 +1,26 @@
+package grafeasos
+
+import (
+	grafeaspb "github.com/grafeas/client-go/0.1.0"
+)
+
+const noNewOcc = "no new occurrences generated"
+
+// Item implements a MetadataItem.
+type Item struct {
```
catherinejones: I think Item might be a relic of a bygone era! We should discuss getting rid of this function!

grafeasos/client.go
```diff
@@ -0,0 +1,221 @@
...
+		for {
+			for _, note := range notesResponse.Notes {
+				if fn(&note, noteKind) {
+					optsOccurrences := &grafeaspb.ListNoteOccurrencesOpts{PageSize: optional.NewInt32(5)}
+					occsResponse, _, errOcc := g.grafeas.ListNoteOccurrences(ctx, note.Name, optsOccurrences)
+					if errOcc != nil {
+						err = errOcc
+						continue
+					}
+					for {
```
catherinejones: As I mentioned in the containeranalysis API part of this PR, could we split up this function into two?

containeranalysis/error.go
```diff
@@ -1,12 +1,12 @@
...
+package containeranalysis
 
 import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
-// isAttestationExistsErr returns true if the passed Error is an "AlreadyExists" gRPC error.
-func isAttestionExistsErr(err error) bool {
+// IsAttestationExistsErr returns true if the passed Error is an "AlreadyExists" gRPC error.
```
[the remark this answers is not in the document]
lynnsh: It is used by the grafeasos package. I thought there was no change in logic here for grafeasos, so I could reuse it instead. Would something else work better here?

grafeasos/client.go
lynnsh: But `grafeasos/client.go` should implement MetadataClient interface that requires the method `GetMetadata()`?

catherinejones: Is there anything that is still using GetMetadata externally? Everything is updated to use the `GetVulnerabilities` functions and so on that you added

containeranalysis/error.go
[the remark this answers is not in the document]
catherinejones: Could we duplicate it? I think this is a reasonable example of where it makes sense to repeat ourselves, since we might get a different output from grafeas if it moves forward faster than containeranalysis and vis-versa!

grafeasos/client.go
lynnsh: Except for polling nothing else uses it.

containeranalysis/error.go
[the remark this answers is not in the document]
lynnsh: Ok, I will do that.

grafeasos/client.go
catherinejones: Cool! Could we move that logic into polling and simplify it?

lynnsh: Yes, but since polling has access only to MetadataClient, but not the underlying client we would need to either pass the actual client to polling or move polling to `client.go`?

catherinejones: hmm what do you think makes more sense? I'm not opposed to polling moving into `client.go`!

lynnsh: Both options seem fine? `polling` is a private function, so it doesn't matter what kind of client is passed to it. And since it is a part of GetVulnerabilities, it is fine to move it to the client. So it's more about efficiency (copying the client when passing to another function) vs clarity (the client file is not too big)?

catherinejones: Since we pass the client by reference (using a pointer) there shouldn't be any speed impact over using the interface! :> One compromise might be moving `GetMetadata` to a new function that only polls for discoveries? And then passing the client as a pointer to the polling function!

lynnsh: So having a separate function for polling that `poll.go` would use?

catherinejones: yeah exactly! You could literally just rename `GetMetadata` and clean it up a bit :)

lynnsh: Ok! I would still have to have GetMetadata to comply with the interface unless we want to remove this function from everywhere? (then polling in containeranalysis would need to be updated as well)

catherinejones: Let's remove it from everywhere! :D We can update the MetadataClient interface and update the containeranalysis package as well!

lynnsh: Ok, I will do that! :)

grafeasos/item.go
lynnsh: Should the whole struct be removed? Does it mean the wrapper is not necessary or should it have a different structure?

catherinejones: Where are we using Item? I think we're just using it in `GetMetadata`, right?

lynnsh: I think by extension I used it in some other methods too, but it can be removed. Same goes for containeranalysis, right?

catherinejones: Yes let’s remove it in both!

lynnsh: The issue with removing `Item` from grafeasos and containeranalysis is this function: `AddAttestationToImage(context.Context, ImageData, AttestationPayload) (MetadataItem, error)`. It returns `MetadataItem` that is an interface that should have a String() method. But grafeasos API `Occurrence` does not have a string method.

catherinejones: Let's leave `Item` for now and replace it later!

cmd/config/metadataclient.go
```diff
@@ -18,7 +19,17 @@ func NewMetadataClient(ctx context.Context) (voucher.MetadataClient, error) {
 		keyring = nil
 	}
 
-	return grafeas.NewClient(
+	isGrafeasosClient := viper.GetBool("metadata_client_grafeasos")
```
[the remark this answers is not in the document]
lynnsh: Done.

containeranalysis/client.go
```diff
@@ -48,20 +48,32 @@ func (g *Client) GetMetadata(ctx context.Context, reference reference.Canonical,
...
 		if nil != err {
 			if iterator.Done == err {
 				err = nil
 			}
 			break
 		}
-		item := new(Item)
-		item.Occurrence = occ
-		items = append(items, item)
+		reqOccurrences := &containeranalysis.ListNoteOccurrencesRequest{Name: note.Name, Filter: filterStr}
```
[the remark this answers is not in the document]
lynnsh: Done.

containeranalysis/error.go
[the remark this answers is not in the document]
lynnsh: Done.

grafeasos/client.go
lynnsh: Done.

grafeasos/client.go
lynnsh: Done.

scanner/scanner.go
lynnsh: Done.

grafeasos/client.go
lynnsh: Done.

cmd/config/metadataclient.go
```diff
@@ -18,10 +21,15 @@ func NewMetadataClient(ctx context.Context) (voucher.MetadataClient, error) {
...
-	return grafeas.NewClient(
-		ctx,
-		viper.GetString("image_project"),
-		viper.GetString("binauth_project"),
-		keyring,
-	)
+	metadataClient := viper.GetString("metadata_client")
+	switch metadataClient {
+	case "grafeasos":
+		return grafeasos.NewClient(ctx, viper.GetString("image_project"),
```
catherinejones: Could you add a flag for us to be able to set the `hostname` and the `port` of the metadata server instance?

cmd/config/scanner.go
```diff
@@ -22,7 +21,7 @@ func newScanner(metadataClient voucher.MetadataClient, auth voucher.Auth) (scann
 		}
 		scanner = clair.NewScanner(config, auth)
 	case "gca", "g":
```
catherinejones: We should deprecate this option:

```golang
case "gca", "g":
      log.Warningf("the %s option for `scanner` has been deprecated and will be removed in the future. Please use `metadata` instead.", scannerName)
      scanner = voucher.NewScanner(metadataClient)
case "metadata":
      scanner = voucher.NewScanner(metadataClient)
```

cmd/config/metadataclient.go
lynnsh: Done.

cmd/config/scanner.go
lynnsh: Done.

catherinejones:
Closing this PR, as I split it up into a GA release version. I'll do a followup PR to merge @lynnsh's work on getting the Grafeas client working!

PR #70, opened 2020-01-24
update to ContainerAnalysis GA release

catherinejones:
This PR takes the work we did to support the open source Grafeas implementation, and focuses it on updating Voucher's support for the ContainerAnalysis API GA release.


cmd/config/metadataclient.go
```diff
@@ -18,10 +18,22 @@ func NewMetadataClient(ctx context.Context) (voucher.MetadataClient, error) {
...
 	}
 
-	return grafeas.NewClient(
-		ctx,
-		viper.GetString("image_project"),
-		viper.GetString("binauth_project"),
-		keyring,
-	)
+	metadataClient := viper.GetString("metadata_client")
+	switch metadataClient {
```
Nilay-Shah: Why have a case statement If we only support 1 option? I'm thinking maybe we have an `if` statement which will do the `log.Warning` message and continue returning a `containeranalysis.NewClient` anyways?

testing/metadata_client.go
```diff
@@ -68,9 +64,8 @@ func (c *MetadataClient) GetBuildDetails(ctx context.Context, reference referenc
 	return c.details[refString], nil
 }
 
-func (c *MetadataClient) AddAttestationToImage(ctx context.Context, reference reference.Canonical, payload voucher.AttestationPayload) (voucher.MetadataItem, error) {
-	occ := new(voucher.MetadataItem)
-	return *occ, nil
+func (c *MetadataClient) AddAttestationToImage(ctx context.Context, reference reference.Canonical, payload voucher.AttestationPayload) (interface{}, error) {
+	return nil, nil
```
Nilay-Shah: I'm unsure what this does/provide?

cmd/config/metadataclient.go
catherinejones: We will be supporting more options once the open source Grafeas PR is finished! Originally this switch statement had both grafeas and containeranalysis as options.

testing/metadata_client.go
catherinejones: This ensures that it implements voucher.MetadataClient, which requires this function. We don't have tests which utilize this stub, so we just leave it as empty.

PR #72, opened 2020-01-29
Add Approved check

genevieveluyt:
Adds a repo check! The check verifies that the changes made to an application (and thus the changes added to the container image) passed the appropriate checks set out in the source code repo.

It does this by doing the following:
1. Call to Github API, get the default branch (usually `master` but may be something else)
1. If the image was built off of the default branch, check that the last commit was signed (if not the default branch, the check fails, which means the image is not signed with this change)
1. If the last commit was signed and is a merge commit, check the PR (if the last commit was not associated with a pull request somehow, the check fails)
1. If that PR had the right number of signatures and checks, the Voucher check passes


checks/README.md
```diff
@@ -36,7 +36,7 @@ Below is the Check interface from the source code.
 
 ```golang
 type Check interface {
-	Check(ImageData) (bool, error)
+	Check(context.Context, ImageData) (bool, error)
```
catherinejones: Thank you for fixing this documentation! Could I request that you move this into a separate PR so we can merge it ASAP?

genevieveluyt: https://github.com/Shopify/voucher/pull/73

catherinejones:
After some reviewing and reading up on the `testify/mocks` library, I'm going to go out on a limb here and say "lets use testify/mock" mocks! Please delete the original mocks that you've already re-created!

catherinejones:
Could you move the mock refactoring into a separate PR?

checks/repo/check.go
```diff
@@ -0,0 +1,94 @@
...
+		if pullRequest.IsMerged && pullRequest.MergeCommit.URL == commit.URL {
+			return true
+		}
+	}
+	return false
+}
+
+// passedCI checks that all Github CI checks have completed and passed
+func passedCI(commit repository.Commit) bool {
+	return commit.Status == "SUCCESS"
```
catherinejones: Could we make `SUCCESS` a constant?

checks/repo/check.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func init() {
+	voucher.RegisterCheckFactory("repo", func() voucher.Check {
```
catherinejones: I think here we should use "approved" or something else to reflect that we're verifying that the check was appropriately approved! I think "repo" is too vague...

checks/repo/check.go
```diff
@@ -0,0 +1,94 @@
...
+
+// isFromBranch checks that the commit is the most recent commit on the branch
+func isFromBranch(branch repository.Branch, commit repository.Commit) bool {
+	latestCommitRef := branch.CommitRefs[0].URL
+	return commit.URL == latestCommitRef
+}
+
+// isSigned checks that the commit is signed
+func isSigned(commit repository.Commit) bool {
+	return commit.IsSigned
```
catherinejones: Will this return true if Github doesn't trust the signing key?

cmd/config/repoclient.go
```diff
@@ -13,57 +11,59 @@ import (
...
-	if key, ok := findLongestKeyMatch(keyring, orgURL); ok {
-		token := keyring[key]
+	orgs := GetOrganizationsFromConfig()
+	if alias, ok := getOrgAlias(orgs, org); ok {
+		token := keyring[alias]
 		return &token, nil
 	}
 
-	return nil, fmt.Errorf("failed to get token for %s: %w", orgURL, err)
+	return nil, fmt.Errorf("failed to get token for %s/%s: %w", org.VCS, org.Name, err)
```
catherinejones: Should `org.Name` here be the alias?

cmd/config/get_orgs_config.go
```diff
@@ -15,9 +15,9 @@ func GetOrganizationsFromConfig() (orgs map[string]repository.Organization) {
 	}
 	for _, row := range repositories {
 		if m, ok := row.(map[string]interface{}); ok {
-			name := m["org-name"].(string)
+			alias := m["alias"].(string)
```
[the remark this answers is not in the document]
genevieveluyt: I was under the impression that we don't need the `org-name` field in the config anymore since we can parse it out of the url, is that right? I agree that there should be a check here to make sure it's not empty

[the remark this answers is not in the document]
catherinejones: Oh, good call! Should we throw an error if `alias` is empty?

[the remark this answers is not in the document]
genevieveluyt: `repository.NewOrganization(alias, url)` parses out the org name from the URL and sets alias to that if it's empty so we could do
```
org := *repository.NewOrganization(alias, url)
orgs[org.Alias] = org
```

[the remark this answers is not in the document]
catherinejones: Oh, yes, that'd be a good idea! :>

What do we do if multiple orgs have the same alias?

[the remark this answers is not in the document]
genevieveluyt: Hm maybe if the alias is empty we should set it to the longest possible value (ie. org + repo) to minimize this risk?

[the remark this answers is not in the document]
genevieveluyt: We could also prefix or postfix it with something so it won't conflict with aliases the user has set explicitly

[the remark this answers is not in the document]
catherinejones: That's true!

As an aside, should we change the config syntax to be more inline with the `required.env` syntax for the Check groups?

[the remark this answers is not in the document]
genevieveluyt: Something like
```
[repository.my-alias]
url = "https://..."
```
? I think that does look nicer. Plus if we can enforce the alias not being empty in the config I think that would be ideal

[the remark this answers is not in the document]
catherinejones: I agree! :shipit:!

checks/repo/check.go
genevieveluyt: The API description of the field is "True if the signature is valid and verified by GitHub." (ref we populated `isSigned` using the `isValid` field https://developer.github.com/v4/interface/gitsignature/) so I think no?

catherinejones: Good enough for me! :D

checks/approved/check_test.go
```diff
@@ -0,0 +1,98 @@
...
+	r "github.com/Shopify/voucher/repository"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRepoCheck(t *testing.T) {
+	ctx := context.Background()
+	imageData, err := voucher.NewImageData("gcr.io/voucher-test-project/apps/staging/voucher-internal@sha256:73d506a23331fce5cb6f49bfb4c27450d2ef4878efce89f03a46b27372a88430")
+	require.NoErrorf(t, err, "failed to get ImageData: %s", err)
+	buildDetail := r.BuildDetail{RepositoryURL: "https://github.com/Shopify/app", Commit: "efgh6543"}
```
catherinejones: Should this be updated to `Shopify/voucher`?

repository/objects.go
```diff
@@ -1,19 +1,61 @@
...
+				VCS:   match["vcs"],
+				Name:  match["org"],
+			}
+		}
 	}
+	return nil
 }
 
+const (
+	CommitStatusError    = "ERROR"
```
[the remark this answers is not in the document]
genevieveluyt: Currently they are but I think if we add more repository types we could map whatever statuses they have to these ones?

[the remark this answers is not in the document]
catherinejones: Sounds good! 👍

checks/approved/check_test.go
genevieveluyt: It doesn't really matter since the mock responses don't look at the actual url. If we wanted to be consistent it should probably be `Shopify/voucher-internal` to match the image data?

catherinejones: Could we update the URLs just to be consistent? Very nitpicky, I know 😅

genevieveluyt: Sure 🙂

PR #74, opened 2020-01-31
Refactor PGP signer into a separate module

actgardner:
Pull the PGP keyring and signing logic into their own module behind the `AttestationSigner` interface. This is the first step to adding support for PKIX and KMS signers.


signer/pgp/keyring.go
```diff
@@ -20,6 +20,16 @@ type KeyRing struct {
 	entities openpgp.EntityList
 }
 
+func (keyring *KeyRing) Sign(checkName, body string) (string, string, error) {
```
catherinejones: Should we change body to a `[]byte`?

PR #75, opened 2020-01-31
return error code when checks fail in voucher_client

sudojimmy:


cmd/voucher_client/submit.go
```diff
@@ -49,6 +49,10 @@ func submit(ctx context.Context, client voucher.Interface, check string, canonic
 
 	fmt.Println(formatResponse(&voucherResp))
 
+	if !voucherResp.Success {
+		return fmt.Errorf("image checking failed")
```
catherinejones: Could you create a new error type?

```golang
import (
    "errors"
)

var errImageCheckFailed = errors.New("image failed to pass required check(s)")
```

PR #76, opened 2020-02-03
Add Google Cloud KMS signer

actgardner:
Adds Google Cloud KMS as a signing method for attestations:
- add a new configuration table for KMS keys
- prefer KMS keys to PGP keys for a given check if both are configured, since PGP is going away in favour of PKIX
- only support SHA512 digests for now for KMS keys


containeranalysis/client.go
```diff
@@ -22,14 +22,15 @@ var errCannotAttest = errors.New("cannot create attestations, keyring is empty")
...
+	kmsKeyring        signer.AttestationSigner // The KMS keys used for signing metadata. KMS keys are preferred to PGP keys
+	pgpKeyring        signer.AttestationSigner // The PGP keyring used for signing metadata. PGP keys are used only if no KMS key is configured for a check
 	binauthProject    string                   // The project that Binauth Notes and Occurrences are written to.
 	imageProject      string                   // The project that image information is stored.
 }
 
 // CanAttest returns true if the client can create and sign attestations.
 func (g *Client) CanAttest() bool {
-	return nil != g.keyring
+	return nil != g.pgpKeyring || nil != g.kmsKeyring
```
klautcomputing:
```suggestion
	return g.pgpKeyring != g.kmsKeyring
```
⛳️

actgardner: I feel like this is so golfy it becomes unclear ;)

cmd/config/metadataclient.go
```diff
@@ -8,14 +8,29 @@ import (
...
+			keyring = nil
+		}
+	} else if signerName == "kms" {
+		keyring, err = getKMSKeyRing()
+		if nil != err {
+			log.Println("could not load KMS keyring from config, continuing without attestation support: ", err)
+			keyring = nil
+		}
+	} else {
+		log.Printf("Signer %q is unknown, upported values are 'kms' or 'pgp'\n", signerName)
```
catherinejones:
```suggestion
		log.Printf("signer %q is unknown, supported values are 'kms' or 'pgp'\n", signerName)
```

PR #77, opened 2020-02-03
return single build detail

genevieveluyt:


containeranalysis/client.go
```diff
@@ -130,34 +129,22 @@ func (g *Client) GetBuildDetails(ctx context.Context, reference reference.Canoni
...
-		err = &voucher.NoMetadataError{
-			Type: voucher.VulnerabilityType,
-			Err:  errNoOccurrences,
-		}
-	}
-
-	if err != nil {
-		return nil, err
+	if _, err := occIterator.Next(); err != iterator.Done {
+		log.Warn("Found multiple Grafeas occurrences for " + reference.String())
```
catherinejones: Rather than logging here, could we return an error?

PR #79, opened 2020-02-03
Support check groups

genevieveluyt:
Adds support for check groups.

You can now add a configuration block for `required.[env]` to specify which checks should run for the `env` test. The content of this block works the same way as the `checks` block.

For example:
```toml
[required.myenv]
diy = true
provenance = true

[required.myotherenv]
nobody = true
```

This would run the `diy` and `provenance` checks when running the `myenv` test, and the `nobody` check when running the `myotherenv` test.


server/server_test.go
```diff
@@ -80,7 +80,7 @@ func TestInvalidJSON(t *testing.T) {
 
 	for _, route := range Routes {
 		path := route.Path
-		if "/{check}" == path {
+		if "/{name}" == path {
```
catherinejones: Could we add a test for the Check group?

server/routes.go
```diff
@@ -38,10 +38,10 @@ var Routes = []Route{
 		HandleAll,
 	},
 	{
-		"Individual Check",
+		"Individual Check or Check Group",
```
catherinejones: Could we programatically add all of the check groups as Routes?

server/handlers.go
```diff
@@ -92,22 +92,35 @@ func HandleAll(w http.ResponseWriter, r *http.Request) {
...
 
-	if voucher.IsCheckFactoryRegistered(checkName) {
-		handleChecks(w, r, checkName)
+	if voucher.IsCheckFactoryRegistered(name) {
+		handleChecks(w, r, name)
 		return
 	}
 
-	http.Error(w, fmt.Sprintf("check %s is not available", checkName), 404)
+	requiredChecks := config.GetRequiredChecksFromConfig()
```
catherinejones: Could we move this into it's own function, such as `HandleCheckGroup`?

PR #80, opened 2020-02-04
Refactor mocks

genevieveluyt:


mock_repository_client.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+
+	"github.com/stretchr/testify/mock"
+
+	r "github.com/Shopify/voucher/repository"
+)
+
+type MockRepositoryClient struct {
```
catherinejones: Could this move into `repository.MockClient`?

catherinejones:
It looks good to me but I think the the tests are broken now, when you have a minute could you investigate? 😅

genevieveluyt:
@catherinejones As per our conversation, I think the testing of the headers is covered by `TestAddPreviewSchemaHeaders` so we can just remove the "flaky" test. Let me know if you disagree

PR #81, opened 2020-02-06
Add support for Github Installation credentials

actgardner:
Add the option to configure Github access using Github Applications instead of personal access tokens.


actgardner:
Tested internally, I was able to produce `is_shopify` attestations in staging.

PR #82, opened 2020-02-07
Load EJSON data at startup

actgardner:
Currently we call readEjson up to twice per request, to load Github credentials and to load the PGP keyring (if the PGP signer is used). This PR would load and decrypt the EJSON data once at startup to reduce the work we do during a request.

If we're interested in this kind of change we could also:
- cache some of the clients
- validate config on startup and fail fast, to simplify the request handling logic


cmd/config/ejson.go
```diff
@@ -13,76 +13,43 @@ import (
 	"github.com/Shopify/voucher/signer/pgp"
 )
 
-// ejsonFormat represents the format that the ejson configuration is structured
-// in.
-type ejsonFormat struct {
+// ejsonData represents the format that the ejson configuration is structured in.
+type EJSONData struct {
```
catherinejones: Could we rename this to `config.Secrets` or something to that effect? EJSONData is a mouthful and if we're going to export it I think it should be clearer why this is isn't a part of the other configuration structures.

cmd/config/ejson.go
```diff
@@ -13,76 +13,43 @@ import (
...
 	Keys                     map[string]string  `json:"openpgpkeys"`
 	ClairConfig              clair.Config       `json:"clair"`
 	RepositoryAuthentication repository.KeyRing `json:"repositories"`
 }
 
 // readEjson reads from the ejson file and populates the passed interface.
-func readEjson(data interface{}) error {
+func ReadEjson() (*EJSONData, error) {
```
catherinejones: Could we change this to `ReadSecrets` or something, so that we're consistent with the renaming of `EJSONData`?

cmd/config/repoclient.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // NewRepositoryClient creates a new repository.Client for the given repository URL. The URL may be in any known
 // format including, but not limited to, urls starting with 'http://', 'https://', 'git@', etc.
-func NewRepositoryClient(ctx context.Context, repoURL string) (repository.Client, error) {
+func NewRepositoryClient(ctx context.Context, ejson *EJSONData, repoURL string) (repository.Client, error) {
```
catherinejones: Could we use the `repository.KeyRing` structure instead of the full `EJSONData` structure here?

actgardner:
Rebased and fixed up the feedback, ready for a second round.

cmd/voucher_server/server.go
```diff
@@ -13,15 +13,20 @@ var serverCmd = &cobra.Command{
...
 			Port:        viper.GetInt("server.port"),
 			Timeout:     viper.GetInt("server.timeout"),
 			RequireAuth: viper.GetBool("server.require_auth"),
 			Username:    viper.GetString("server.username"),
 			PassHash:    viper.GetString("server.password"),
 		}
-		server.Serve(&config)
+		secrets, err := config.ReadSecrets()
+		if err != nil {
+			return err
```
catherinejones: Could we log this error rather than fail on it? We could foreseeably run Voucher without secrets.

cmd/voucher_server/server.go
```diff
@@ -14,14 +16,18 @@ var serverCmd = &cobra.Command{
...
 			RequireAuth: viper.GetBool("server.require_auth"),
 			Username:    viper.GetString("server.username"),
 			PassHash:    viper.GetString("server.password"),
 		}
-		server.Serve(&config)
+		secrets, err := config.ReadSecrets()
+		if err != nil {
+			log.Printf("Error loading EJSON file, no secrets loaded: %v", err)
+		}
+		server.Serve(&serverConfig, secrets)
```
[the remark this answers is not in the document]
actgardner: I went back and added cases everywhere so we log an error if we need unconfigured secrets. Long term I'd love to pull all of that stuff into the server initialization so we're not ad-hoc validating config on every request.

PR #83, opened 2020-02-14
Add support for metrics with DogStatsd

actgardner:
Add a generic metric client with an implementation backed by datadog's statsd client.


config/secrets.ejson
```diff
@@ -0,0 +1,15 @@
+{
```
catherinejones: Is this example data?

actgardner: Whoops yeah removed.

PR #84, opened 2020-02-14
Actually check for approvals

genevieveluyt:
As it turns out the commit status returned by Github does not account for required approvals (if the branch requires approvals but the PR is merged without them, the commit "status" is still "SUCCESS"). This checks that the PR actually meets the approvals requirements by comparing the number of PR reviews that have state "APPROVED" to the branch protection requirements for the base branch of the PR.


repository/github/queries.go
```diff
@@ -77,15 +77,89 @@ func newCommitInfoResult(ctx context.Context, ghc ghGraphQLClient, commitURL str
 		return repository.Commit{}, err
 	}
 	queryResult := new(commitInfoQuery)
+
```
catherinejones: This file is terrifying to review, could you split it into multiple files based on query? 😅

checks/approved/check.go
```diff
@@ -75,14 +90,17 @@ func isSigned(commit repository.Commit) bool {
 	return commit.IsSigned
 }
 
-// isMergeCommit checks that the commit is a merge commit
-func isMergeCommit(commit repository.Commit) bool {
+// isApprovedMergeCommit checks that the commit is a merge commit
+func isApprovedMergeCommit(commit repository.Commit, branch repository.Branch) (passed bool, reason error) {
```
catherinejones: Where is `branch` being used here?

genevieveluyt: Oops missed this when resolving conflicts, good catch!

genevieveluyt:
Re `createHandler`, I didn't write that function but I don't see any reason to keep the unused parameter...

PR #86, opened 2020-02-25
add verify api call

catherinejones:
This PR adds a `verify` API call to Voucher Server and a `verify` option to Voucher Client, which allows users to verify if an image has all required attestations.


catherinejones:
I have yet to write documentation but I will now!

catherinejones:
I'm sorry for all of the force pushing! 😅

cmd/voucher_client/submit.go
```diff
@@ -78,9 +51,9 @@ func LookupAndSubmit(args []string) {
 		os.Exit(1)
 	}
 
-	err = submit(ctx, client, getCheck(), canonicalRef)
+	err = check(ctx, client, getCheck(), canonicalRef)
 	if nil != err {
-		errorf("submitting image to voucher failed: %s", err)
+		errorf("check image to voucher failed: %s", err)
```
genevieveluyt: Maybe `checking image with voucher failed`?

containeranalysis/create.go
```diff
@@ -1,23 +1,36 @@
...
-				PublicKeyId: keyID,
+				Signature:   []byte(attestation.Signature),
+				PublicKeyId: attestation.KeyID,
 			},
 		},
 	}
 
-	occurrenceAttestation := grafeas.Occurrence_Attestation{
-		Attestation: &newAttestation,
+	binauthProjectPath := "projects/" + binauthProject
```
genevieveluyt:
```suggestion
	binauthProjectPath := projectPath(binauthProject)
```

server/server.go
```diff
@@ -34,6 +34,9 @@ func (server *Server) Serve() {
 
 // SetCheckGroup adds a list of checks as a group with the passed name.
 func (server *Server) SetCheckGroup(name string, checkNames []string) {
+	if _, ok := server.checkGroups[name]; !ok {
+		server.checkGroups[name] = make([]string, 0, len(checkNames))
```
genevieveluyt: Just curious what this is contributing? Should we check if `sever.checkGroups` is nil and initialize the map if so?

cmd/voucher_client/README.md
```diff
@@ -64,8 +65,11 @@ $ voucher_client -v http://localhost:8000 gcr.io/path/to/image:latest
...
 ```json
- - Attesting image: gcr.io/path/to/image@sha256:ab7524b7375fbf09b3784f0bbd9cb2505700dd05e03ce5f5e6d262bf2f5ac51c
-   ✗ nobody failed
-   ✗ snakeoil failed, err: vulnernable to 1 vulnerabilities: CVE-2018-12345 (high)
-   ✓ diy succeeded, but wasn't attested, err: rpc error: code = AlreadyExists desc = Requested entity already exists
+Using config file: /Users/catherine/.voucher.yaml
+Submitting image to Voucher: gcr.io/path/to/image@sha256:ab7524b7375fbf09b3784f0bbd9cb2505700dd05e03ce5f5e6d262bf2f5ac51c
+image was rejected
+   ✓ passed nobody
+   ✗ failed snakeoil
```
genevieveluyt: Is it intentional that there's no explanation for the failure anymore? Or just this image doesn't have the same reason for failing than the image used previously?

attestation.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/Shopify/voucher/signer"
+)
+
+// Attestation is a structure that contains the Attestation data that we want
+// to create an MetadataItem from.
+type Attestation struct {
+	CheckName string
+	Body      string
+	Signature string
+	KeyID     string
```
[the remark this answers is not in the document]
catherinejones: Since `SignedAttestation` contains `Attestation`, they are! They can be accessed normally, except when initializing the structure!

https://github.com/Shopify/voucher/pull/86/files#diff-8070e6cb64c127657ffab3c794ddacb5R28-R32

attestation.go
```diff
@@ -0,0 +1,61 @@
...
+// SignAttestation takes a keyring and attestation and signs the body of the
+// payload with it, updating the Attestation's Signature field.
+func SignAttestation(s signer.AttestationSigner, attestation Attestation) (SignedAttestation, error) {
+	signature, keyID, err := s.Sign(attestation.CheckName, attestation.Body)
+	if nil != err {
+		return SignedAttestation{}, err
+	}
+
+	return SignedAttestation{
+		Attestation: Attestation{
```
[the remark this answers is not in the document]
catherinejones: Good point, it likely could be simplified to work like that!

cmd/voucher_client/README.md
catherinejones: Oh, this is because I was getting the output of the verified command, I'll fix this!

attestation.go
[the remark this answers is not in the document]
genevieveluyt: Ah sorry I guess my actual question was, why include those fields in `Attestation` at all?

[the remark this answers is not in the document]
catherinejones: Oh! They shouldn't have been, I think that happened when I was rebasing, sorry! 😅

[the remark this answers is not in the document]
genevieveluyt: Np!

server/server.go
catherinejones: Without doing this, initializing the map doesn't initialize the slice, which makes it behave weirdly when you append things to it. I had a bug where some check names would be called multiple times despite only being added once when this function was called!

PR #88, opened 2020-02-27
Adds Grafeas (open source) MetadataClient

Nilay-Shah:
I used https://github.com/Shopify/voucher/pull/63/ as the base and then rebased on master (and fixed conflicts), remove `/vendor`, ran `make lint-new` (and fixed errors), and retested it (to find a bug in the client-go implementation; more details on this below).
The error is happening because it seems when Swagger is auto generating the code, the URL for the local instance is incorrect; it [doesn't string replace everything properly](https://github.com/grafeas/client-go/blob/master/0.1.0/api_grafeas_v1_beta1.go#L48). Downgrading to the only other version available (`v1alpha1`) won't work as there are no docker images available for this.
For now, we will have to wait until this is resolved (which is why this PR is in draft mode) - https://github.com/grafeas/client-go/pull/15 may likely help with this.

Closes https://github.com/Shopify/voucher/issues/59


catherinejones:
> Downgrading to the only other version available (v1alpha1) won't work as there are no docker images available for this.

Is the v0.1.0 image not able to accept v1alpha1 API calls?

Nilay-Shah:
> Is the v0.1.0 image not able to accept v1alpha1 API calls?
Unfortunately not @catherinejones :(

catherinejones:
Let's hold off on this PR for the time being then! Could you rename this to "adds grafeas MetadataClient"?

PR #89, opened 2020-02-27
refactor API calls

catherinejones:
This PR refactors how the server's API calls work, to remove the need for distinct API calls for groups and checks.


server/handlers.go
```diff
@@ -1,104 +1,24 @@
...
+// all of the Checks in one CheckGroup and creates any attestations if
+// applicable.
+func (s *Server) HandleCheckImage(w http.ResponseWriter, r *http.Request) {
 	var err error
 
-	defer r.Body.Close()
-
 	if err = s.isAuthorized(r); nil != err {
-		http.Error(w, "username or password is incorrect", http.StatusUnauthorized)
+		http.Error(w, "username or password is incorrect", 401)
```
actgardner: This should still use the http status code

catherinejones: Oh yes, good call!

PR #97, opened 2020-06-25
add support for v1 manifests

catherinejones:
This PR adds support for Schema 1 manifests (specifically the signed variant supported by the Docker Distribution library).


testing/manifests.go
```diff
@@ -0,0 +1,136 @@
...
+	"parent":"",
+	"throwaway":true
+}`,
+			},
+		},
+	}
+
+	return manifest
+}
+
```
dani-santos-code, line 115: Oh, cool! It makes sense to move this function here! :)

dani-santos-code:
Can we also remove the `config_call.go` flle since it's empty?

catherinejones:
> Can we also remove the `config_call.go` file since it's empty?

Absolutely, I forgot it was there!

docker/digest.go
```diff
@@ -42,3 +44,36 @@ func GetDigestFromTagged(client *http.Client, image reference.NamedTagged) (dige
...
+	resp, err := client.Do(request)
+	if err != nil {
+		return blank, err
+	}
+
+	if resp.StatusCode >= 300 {
+		return blank, responseToError(resp)
+	}
+
+	_ = resp.Body.Close()
```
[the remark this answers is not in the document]
catherinejones: This is to prevent linters from complaining about us not handling the error on `resp.Body.Close()` (because the error is explicitly ignored). While defer is a reasonable thing to do, it also results in the error not being handled, and it waits to close the body after the function returns (whereas here we're just ignoring it!).

docker/manifest.go
```diff
@@ -1,26 +1,38 @@
...
 
 	request.Header.Add("Accept", schema2.MediaTypeManifest)
+	request.Header.Add("Accept", schema1.MediaTypeManifest)
+	request.Header.Add("Accept", schema1.MediaTypeSignedManifest)
 
-	err = doDockerCall(client, request, &manifest)
+	manifest, err = getDockerManifest(client, request)
+	if nil != err {
+		fmt.Println("failed in RequestManifest to getDockerManifest:", err)
+		return nil, err
```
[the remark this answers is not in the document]
catherinejones: Opps! Those were for me troubleshooting, I've gotten rid of them 😅

docker/digest.go
```diff
@@ -42,3 +44,36 @@ func GetDigestFromTagged(client *http.Client, image reference.NamedTagged) (dige
...
+
+	imageDigest := digest.Digest(resp.Header.Get("Docker-Content-Digest"))
+	if "" == string(imageDigest) {
```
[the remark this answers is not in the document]
catherinejones: It is, but for whatever reason Go doesn't allow comparisons between two different types even if one type is the other type (at least when I wrote this code originally)!

docker/schema1/config.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type v1Blob struct {
+	Config dockerTypes.ExecConfig `json:"config"`
+}
+
+// RequestConfig retrieves the manifest from the associated configuration.
+// Unlike in v2 manifests, v1 manifests have the configuration stored in the
+// history, so we can safely ignore the http.Client passed to this function.
+func RequestConfig(_ *http.Client, _ reference.Canonical, manifest distribution.Manifest) (*dockerTypes.ExecConfig, error) {
```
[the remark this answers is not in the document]
catherinejones: This is so that we can make a generic interface which both `schema1.RequestConfig` and `schema2.RequestConfig` conform to!

For example:

```golang
type RequestConfigFunc func(*http.Client, reference.Canonical, distribution.Manifest) (*dockerTypes.ExecConfig, error)
```

We can then simplify the switch statement to a map of functions:

```golang
map[string]RequestConfigFunc := {
    "schema1mimetype" = schema1.RequestConfig,
    "schema2mimetype" = schema1.RequestConfig,
}
```

I forgot to get to this though

docker/digest.go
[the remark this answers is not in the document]
dani-santos-code: Oh, okie. I saw something similar in `manifest_request.go` (line 18) and was wondering what the diff was. I ended up finding some articles about how "dangerous" this defer can be..since it can return a nil value and break everything. lol
(https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1)

docker/schema1/config.go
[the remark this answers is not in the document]
dani-santos-code: Oh, I see! Thanks for explaining with nice examples! :)

docker/schema1/manifest.go
```diff
@@ -0,0 +1,23 @@
...
+	_, ok := m.(*v1.SignedManifest)
+	return ok
+}
+
+// ToManifest casts a distribution.Manifest to a schema1.Manifest. It panics
+// if it passed anything other than a schema1.SignedManifest.
+func ToManifest(manifest distribution.Manifest) *v1.SignedManifest {
+	signedManifest, ok := manifest.(*v1.SignedManifest)
+	if !ok {
+		panic("schema1.ToManifest was passed a non-schema1.SignedManifest")
```
rxbchen, line 19: why is `panic` used in this case rather than returning an error?

testing/docker.go
```diff
@@ -7,9 +7,10 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/manifest/schema2"
 	dockerTypes "github.com/docker/docker/api/types"
+	"github.com/docker/libtrust"
```
rxbchen, line 13: It seems like this library will be incorporated into `docker/distribution` in the [future](https://github.com/docker/libtrust)

testing/manifests.go
rxbchen: 👍 That's really clean!

docker/digest.go
```diff
@@ -42,3 +44,36 @@ func GetDigestFromTagged(client *http.Client, image reference.NamedTagged) (dige
...
+// image. Returns a digest.Digest, or an error.
+func GetDigestFromCanonical(client *http.Client, image reference.Canonical) (digest.Digest, error) {
+	blank := digest.Digest("")
+
+	request, err := http.NewRequest(http.MethodHead, uri.GetDigestManifestURI(image), nil)
+	if err != nil {
+		return blank, err
+	}
+
+	request.Header.Add("Accept", schema2.MediaTypeManifest)
```
Nilay-Shah, line 58: It seems like everything below this line is the same as lines 25 onwards from the `GetDigestFromTagged` function above. Do you think it's worth creating a helper function and calling that from these 2 methods?

docker/manifest_request.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	return manifest, nil
+}
+
+// isValidManifest ensures that we don't try to unmarshal an invalid manifest.
+func isValidManifest(contentType string) bool {
+	for _, mediaType := range distribution.ManifestMediaTypes() {
+		if mediaType == contentType {
+			return true
```
Nilay-Shah, line 46: This may be my lack of knowledge, so forgive me here! But I find it odd to `return true` if only 1 of many manifest media types are the type we want? Shouldn't we check them all and only `return true` if none of them match?
If we only care about the media type of the manifest list why not just use that instead (since we know there is only 1 of those)?
(what I'm using to understand this structure: https://docs.docker.com/registry/spec/manifest-v2-2/#example-manifest-list)

docker/manifest_test.go
```diff
@@ -6,39 +6,53 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	//"github.com/Shopify/voucher/docker/schema1"
```
Nilay-Shah, line 9: Remove?

clair/scanner.go
```diff
@@ -38,7 +42,13 @@ func (scanner *Scanner) Scan(ctx context.Context, i voucher.ImageData) ([]vouche
 		return vulns, err
 	}
 
-	clairVulns, err := getClairVulnerabilities(manifest, scanner.config, tokenSrc, i)
+	if !schema2.IsManifest(manifest) {
+		return vulns, errUnsupportedManifest
+	}
+
+	s2Manifest := schema2.ToManifest(manifest)
```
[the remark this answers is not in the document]
catherinejones: This is to fail on programmer error as opposed to user error. The `ToManifest` function is strictly for working with manifests that have been verified with the `IsManifest` function in the same package!

testing/docker.go
catherinejones: 😲 That'd be so useful, though I can't see any evidence that that has happened since this README was updated four years ago :<

clair/scanner.go
```diff
@@ -38,7 +42,13 @@ func (scanner *Scanner) Scan(ctx context.Context, i voucher.ImageData) ([]vouche
 		return vulns, err
 	}
 
-	clairVulns, err := getClairVulnerabilities(manifest, scanner.config, tokenSrc, i)
+	if !schema2.IsManifest(manifest) {
```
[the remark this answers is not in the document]
catherinejones: I think rather than a comment an Issue might be more worthwhile? A comment clutters the code base and will likely be forgotten...

docker/digest.go
catherinejones: I think that's a reasonable thing to do down the line, I still want to implement retrying for the Docker API and that would be a perfect opportunity to refactor this!

docker/manifest_request.go
catherinejones: When we call the API against a docker image's manifest URL, it could return multiple different manifest content-types (schema 1 signed, schema 2, etc). We don't know if it's a valid manifest (or an HTML page saying "Rate Limited") until we verify what the content-type is. This function verifies that that content type is one that's supported by the Docker distribution library, so we can complain if it's HTML or something else.

docker/manifest_test.go
catherinejones: ooo good catch!

PR #100, opened 2020-07-22
Support v1 manifests in Clair

### LINKED ISSUE #98: Support v1 manifests in Clair
Currently Clair only supports schema v2 Manifests, but we now support version 1 signed manifests as well.

We should add support for version 1.

dani-santos-code:
#97 added support to version1 signed manifests.

Currently, Clair only supports version2, but it would be nice to have it support v1 signed manifests as well. This PR addresses that issue.

Closes #98


clair/clair.go
```diff
@@ -134,5 +171,29 @@ func getClairVulnerabilities(manifest schema2.Manifest, config Config, tokenSrc
...
+// getClairVulnerabilities gets a map[string]v1.Vulnerability from Clair, so that we can convert
+// them to Voucher Vulnerabilities all at once.
+func getClairVulnerabilities(manifest distribution.Manifest, config Config, tokenSrc oauth2.TokenSource, image reference.Canonical) (map[string]v1.Vulnerability, error) {
+	var vulns map[string]v1.Vulnerability
+
+	var err error
+
+	parent := digest.Digest("")
+
+	if s2.IsManifest(manifest) {
```
catherinejones, line 186: It might be worth having a `switch` statement here so we can log if the Clair API gets something that's not a schema1 or schema2 manifest!

dani-santos-code: Oh, nice! Yeah! I removed the error message regarding schema2. 🙈 But there can be a case when neither of those are valid! I will add a switch!! Thanks for pointing it out! 🙏🏽

dani-santos-code:
Thanks, @catherinejones. I'll try to run it with `docker-compose` and see what's happening with `oauth`

PR #101, opened 2020-07-22
adds multi-project support

catherinejones:
Voucher has been severely limited by how it's been unable to deal with images that don't write their metadata to the same project. This PR adds support for getting metadata from the same project that the image is hosted in.

This is a breaking change and will require that metadata be moved from a single project to the same project as the image itself.

For example, if your image is in `my-cool-project`, metadata should now be stored in `my-cool-project` as well.


catherinejones:
I finally got around to testing this change with new permissions and everything is working properly!

klautcomputing (approved):
🚀 🚢

rxbchen (approved):
💯 🚢 🎟️

PR #104, opened 2020-09-25
MetadataClient for Open Source Grafeas

lynnsh:
PR to resolve the [issue](https://github.com/Shopify/binary-authorization/issues/41).
This will close this PR: https://github.com/Shopify/voucher/pull/88.

- add a separate grafeas directory to store both containeranalysis and grafeas open source
- create rest calls to grafeas open source v1beta1 version used by MetadataClient
- create objects to receive and send data used by rest calls (links to similar client-go objects are included in comments)
- add gomock for grafeas open source client tests
- move files that are used by both containeranalysis and grafeas open source into the parent grafeas folder
- add grafeas open source option to cmd/config MetadataClient setup

Steps to test with a real copy of Grafeas:
- start local grafeas server
- add testing data to the server (either with curl and json or by using GrafeasApiService and createAllOccurrences in client_test)
- test using curl requests or update client_test to use GrafeasApiService configured to call your local grafeas server


grafeas/containeranalysis/poll.go
```diff
@@ -44,7 +46,7 @@ func getVulnerabilityDiscoveries(ctx context.Context, client *grafeasv1.Client,
 	occurrencesIterator := client.ListOccurrences(ctx, reqOccurrences)
 
 	for {
-		var occurrence *grafeas.Occurrence
+		var occurrence *grafeasapi.Occurrence
```
dani-santos-code, line 49: should we use `vgrafeas.Occurrence` here instead, since we have that object created and might be safer/easier to maintain down the road?

Makefile
```diff
@@ -85,3 +85,6 @@ release:
 
 snapshot:
 	$(GORELEASER) --snapshot
+
+mocks:
+	mockgen -source=grafeas/rest/grafeas_service.go -destination=grafeas/rest/mocks/grafeas_service_mock.go package=mocks
```
dani-santos-code, line 90: oh, cool! I didn't know `mockgen` / `gomock`

grafeas/containeranalysis/poll.go
lynnsh: I did not update containeranalysis objects because since it is developed by Google, they might change their server definition, and it could become out-of-sync with our objects. But I'm not sure though if `google.golang.org/genproto/googleapis/grafeas/v1` is more reliable than our objects.

grafeas/rest/grafeas_service.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	occResp := objects.ListOccurrencesResponse{}
+	err = json.Unmarshal(resp, &occResp)
+	if err != nil {
+		return objects.ListOccurrencesResponse{}, err
+	}
+	return occResp, nil
+}
+
+//ListNoteOccurrences https://github.com/grafeas/client-go/blob/39fa98b49d38de3942716c0f58f3505012415470/0.1.0/api_grafeas_v1_beta1.go#L949
```
dani-santos-code, line 162: I guess we don't need this one, right? I don't see much of a diff between this and List Occurrence... 🤔 They both have an `Occurrences` field of `[]V1beta10ccurrence ` and a  `NextPageToken` field...

List Note Occurrences:
https://github.com/grafeas/client-go/blob/39fa98b49d38de3942716c0f58f3505012415470/0.1.0/model_v1beta1_list_note_occurrences_response.go

List Occurrences:
https://github.com/grafeas/client-go/blob/39fa98b49d38de3942716c0f58f3505012415470/0.1.0/model_v1beta1_list_occurrences_response.go

grafeas/containeranalysis/poll.go
dani-santos-code: I see...on the repo they mention it is experimental and that it's subject to change...

https://github.com/googleapis/go-genproto

grafeas/rest/grafeas_service.go
lynnsh: It was used initially in grafeas os PR. If I recall correctly, the problem is that grafeas os cannot do filtering, and so we are getting a lot of occurrences at once (especially for build, where we expect one at most, getting everything seems wasteful). 
If we use `ListNoteOccurrences`, we would get only notes, filter them locally for `kind`, and then get occurrences for all notes that are of the `kind` we are interested in. Technically, it would result in getting less occurrences, but we do need 2 API calls here. But, I tried to test locally `ListNoteOccurrences`, and it does not work for me. It never gets occurrences for a specific note, and the docs seem to be outdated, so I am not sure if this is possible.

grafeas/containeranalysis/poll.go
lynnsh: But if they change it to keep it in sync with containeranalysis, then maybe it is a good thing? Probably I'll need to take a look at proper examples of containeranalysis to see what is used there. If they themselves recommend this repo, then I'll leave it, if not, then maybe it should be changed?

grafeas/rest/client.go
```diff
@@ -0,0 +1,192 @@
...
+
+// GetAttestations returns all of the attestations associated with an image
+func (g *Client) GetAttestations(ctx context.Context, reference reference.Canonical) ([]voucher.SignedAttestation, error) {
+	var attestations []voucher.SignedAttestation
+
+	occurrences, err := g.getAllOccurrences(ctx)
+	if err != nil {
+		return []voucher.SignedAttestation{}, err
+	}
+	for _, occ := range occurrences {
```
dani-santos-code, line 84: here we are not passing a filter any longer (for Vulnerability specifically) as `*objects.ListOpts` to `ListOccurrences`. Wouldn't it make that call to the api a bit faster, rather than having it go through every kind?

Then instead of:	
```
for _, occ := range occurrences {
      if *occ.Kind == objects.NoteKindVulnerability {
		item := objects.OccurrenceToVulnerability(&occ, vulProject)
		items = append(items, item)
}
```

you would go for something like:


```
for _, occ := range occurrences {
	attestations = append(
	  attestations,
	  objects.OccurrenceToAttestation(note, &occ),
	 )
}
```

grafeas/rest/client.go
```diff
@@ -0,0 +1,192 @@
...
+	if nil != err {
+		return []voucher.Vulnerability{}, err
+	}
+
+	occurrences, err := g.getAllOccurrences(ctx)
+	if nil != err {
+		return []voucher.Vulnerability{}, err
+	}
+	for _, occ := range occurrences {
+		if *occ.Kind == objects.NoteKindVulnerability {
```
dani-santos-code, line 116: same here

grafeas/rest/client.go
lynnsh: Filtering does not work in this version of grafeas. It was definitely the case last fall, and from what I've checked now, there were no new updates regarding this.

grafeas/containeranalysis/poll.go
dani-santos-code: you're right! they'll likely update if necessary and we'll be in sync!

grafeas/rest/client.go
dani-santos-code: Oh, I see! That's what you were explaining in the previous comment!

grafeas/rest/grafeas_service.go
dani-santos-code: Oh, I see!! Thanks for explaining!! I'll pull the docs and see if I find anything that could help. Thanks a lot for clarifying! 🙏🏽

grafeas/containeranalysis/poll.go
lynnsh: Judging by how containeranalysis/v1beta1 is constructed, I'd think that they have their own internal objects, and they do not use go-genproto. However, I see that it was recently updated to v1, similar to containeranalysis, so it might be more reliable to use it for now than our own objects? Even if we decide to use our own ones, we probably would need to create something separate from grafeas os in case containeranalysis is changed, since with using the same objects we could break one of the implementations. But I'd argue, if we want to do it, it should be in another issue since it is not that related to grafeas os?

grafeas/rest/client.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	occurrences := []objects.Occurrence{}
+	for _, occ := range items {
+		if *occ.Kind == objects.NoteKindBuild {
+			occurrences = append(occurrences, occ)
+		}
+	}
+
+	// we should only have 1 occurrence based on our kind specified
+	if nil == err && len(occurrences) != 1 {
```
dani-santos-code, line 149: good point!

grafeas/rest/client.go
```diff
@@ -0,0 +1,192 @@
...
+// NewClient creates a new Grafeas Client.
+func NewClient(ctx context.Context, imageProject, binauthProject string, keyring signer.AttestationSigner, grafeas GrafeasAPIService) (*Client, error) {
+	client := new(Client)
+	client.grafeas = grafeas
+	client.keyring = keyring
+	client.binauthProject = binauthProject
+	client.imageProject = imageProject
+
+	return client, nil
+}
```
dani-santos-code, line 192: here you could go for:

`return &Client{grafeas, keyring, binauthProject, imageProject}, nil`

grafeas/rest/client_test.go
```diff
@@ -0,0 +1,367 @@
...
+		expectedError  error
+	}{
+		"valid input": {
+			returnOccs: objects.ListOccurrencesResponse{
+				Occurrences: occs,
+			},
+			expectedResult: []voucher.Vulnerability{{
+				Name:     "notename",
+				Severity: voucher.NegligibleSeverity,
+			}},
```
dani-santos-code, line 213: should we add `expectedError: nil` here?

lynnsh: It is `nil` by default but I can add it to make it more explicit?

grafeas/rest/client.go
lynnsh: Done.

grafeas/rest/poll.go
```diff
@@ -0,0 +1,68 @@
...
+
+	return false
+}
+
+// pollForDiscoveries pauses execution until grafeas has pushed
+// the Vulnerability information to the server.
+func pollForDiscoveries(ctx context.Context, c *Client) error {
+	for i := 0; i < attempts; i++ {
+		discoveries, err := getVulnerabilityDiscoveries(ctx, c)
+		if err != nil && !voucher.IsNoMetadataError(err) {
```
dani-santos-code, line 37: don't we want to check `if err != nil && voucher.IsNoMetadataError(err)` ?

Say the list of  occurrences is empty ([]objects.Occurrence), you won't find any discoveries and voucher will raise a NoMetadataError, right?

grafeas/rest/client_test.go
dani-santos-code: yeah...or maybe remove the `expectedError` from the struct on line 204, since it's not being used?

dani-santos-code: Could we also test the unhappy path? Something like:
```
		"invalid input": {
			returnOccs:     objects.ListOccurrencesResponse{},
			expectedResult: []voucher.Vulnerability{},
			expectedError: &voucher.NoMetadataError{
				Type: voucher.VulnerabilityType,
				Err:  grafeas.ErrNoOccurrences,
			},
		},
```

then:

`grafeasMock.EXPECT().ListOccurrences(gomock.Any(), gomock.Any(), gomock.Any()).Return(test.returnOccs, test.expectedError).MaxTimes(2)`

grafeas/containeranalysis/poll.go
dani-santos-code: yeah, I agree that they probably don't use `go-genproto`. The last major update to the grafeas objects was 15-17 months ago: https://github.com/googleapis/go-genproto/tree/master/googleapis/grafeas/v1

I guess we can keep the objects you created for the os version for now. And maybe let's see what others in the team think about it?

And you're right. If we want to change the objects in the `containeranalysis`, this should be in a separate PR.

grafeas/rest/poll.go
lynnsh: The idea for polling is to continue to check for data if grafeas has not pushed Vulnerability information yet to the server. When we call `getVulnerabilityDiscoveries` we expect 3 cases: there are vulnerabilities (which means we can return and get them), there are no vulnerabilities and no other errors (meaning that maybe grafeas did not yet push the info we need to the server, so it is fine and we would sleep and check again), and there is some other error (which we return because there is no reason to wait). So in this case we don't want to raise `NoMetadataError` yet because we want to continue polling to check if we get some data eventually.

dani-santos-code: Oh, I see! You're attempting to poll 5 times. Thanks for clarifying!! 🙏🏽

grafeas/rest/client_test.go
lynnsh: It is a problem to set it up with Vulnerability because of polling. If you have no data, then the polling will make the test to time out and return the error even if there are no issues. But I tried the test where there is only a discovery occurrence, so polling would be successful, but without vulnerabilities `NoMetadataError` will return.

grafeas/rest/objects/vulnerability.go
```diff
@@ -0,0 +1,140 @@
+package objects
```
Nilay-Shah, line 1: Should files like this be called `vulnerability_occurrence` instead? Since we have 1 `objects` package that also contains things like `note`, `occurrence`, etc. I feel this extra descriptiveness may help differentiating between all these various objects. What do you think? I may very well be missing something 😅

Nilay-Shah (commented):
Some initial thoughts after parsing the code! Overall the functionality looks okay and so does the moving around of files! I haven't pulled this code to test it out locally, but I'll do that soon.

grafeas/rest/client.go
```diff
@@ -0,0 +1,191 @@
+package rest
+
+import (
+	"context"
+	"errors"
+
+	"github.com/Shopify/voucher"
+	"github.com/Shopify/voucher/attestation"
+	"github.com/Shopify/voucher/grafeas"
```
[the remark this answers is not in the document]
lynnsh: Done.

grafeas/rest/client.go
```diff
@@ -0,0 +1,191 @@
...
+	attestation := objects.AttestationDetails{Attestation: &objects.Attestation{
+		PgpSignedAttestation: &objects.AttestationPgpSigned{Signature: signedAttestation.Signature,
+			PgpKeyID: signedAttestation.KeyID, ContentType: &contentType}}}
+
+	occurrence := objects.NewOccurrence(reference, payload.CheckName, &attestation, binauthProjectPath)
+	_, err = g.grafeas.CreateOccurrence(ctx, binauthProjectPath, occurrence)
+
+	if grafeas.IsAttestationExistsErr(err) {
+		err = nil
+		signedAttestation.Signature = ""
```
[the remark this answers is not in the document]
lynnsh: This is the same logic as in the containeranalysis package, and I thought that it is required to unset the signature if the found attestation already exists, but I am not sure why, to be honest, I could not find this being used anywhere in voucher.

grafeas/rest/client.go
```diff
@@ -0,0 +1,191 @@
...
+// GetAttestations returns all of the attestations associated with an image
+func (g *Client) GetAttestations(ctx context.Context, reference reference.Canonical) ([]voucher.SignedAttestation, error) {
+	var attestations []voucher.SignedAttestation
+
+	occurrences, err := g.getAllOccurrences(ctx)
+	if err != nil {
+		return []voucher.SignedAttestation{}, err
+	}
+	for _, occ := range occurrences {
+		if *occ.Kind == objects.NoteKindAttestation {
```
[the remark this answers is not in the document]
lynnsh: Done.

grafeas/rest/objects/vulnerability.go
lynnsh: Except for exchange, occurrence, and note objects, other files have structs used in both occurrence and note files, so to make it more explicit, I could divide each file, for example, to vulnerability_occurrence and vulnerability_note?

grafeas/containeranalysis/poll.go
catherinejones: Since we're using gRPC for the containeranalysis package and REST for the other package I agree that it makes sense not to use the same objects for both!

grafeas/rest/client.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/Shopify/voucher/signer"
+	"github.com/antihax/optional"
+	"github.com/docker/distribution/reference"
+)
+
+var errCannotAttest = errors.New("cannot create attestations, keyring is empty")
+
+// vulProject is the project that Google's Container Analysis writes vulnerability
+// occurrences to.
+const vulProject = "projects/goog-vulnz/notes/"
```
catherinejones, line 21: I think we need to make this into a configuration variable, as an open source grafeas instance is less likely to follow the Google project names!

grafeas/rest/grafeas_service.go
```diff
@@ -0,0 +1,210 @@
...
+	CreateOccurrence(context.Context, string, objects.Occurrence) (objects.Occurrence, error)
+	BatchCreateOccurrences(context.Context, string, []objects.Occurrence) ([]objects.Occurrence, error)
+	BatchCreateNotes(context.Context, string, map[string]objects.Note) ([]objects.Note, error)
+	ListNotes(context.Context, string, *objects.ListOpts) (objects.ListNotesResponse, error)
+	ListOccurrences(context.Context, string, *objects.ListOpts) (objects.ListOccurrencesResponse, error)
+	//ListNoteOccurrences(context.Context, string, string, *objects.ListOpts) (objects.ListOccurrencesResponse, error)
+}
+
+//GrafeasAPIServiceImpl for REST calls
+type GrafeasAPIServiceImpl struct {
```
catherinejones, line 35: Since we have the `GrafeasAPIService` interface for this, we can unexport this structure so as to avoid people accidentally interacting with it!

grafeas/rest/objects/attestation.go
```diff
@@ -0,0 +1,54 @@
+package objects
+
+//AttestationSignedContentType https://github.com/grafeas/client-go/blob/master/0.1.0/model_attestation_pgp_signed_attestation_content_type.go
```
catherinejones, line 3: How much do these differ between those in the grafeas/client-go package?

grafeas/rest/client.go
```diff
@@ -0,0 +1,193 @@
...
+	signedAttestation, err := voucher.SignAttestation(g.keyring, payload)
+	if nil != err {
+		return voucher.SignedAttestation{}, err
+	}
+
+	binauthProjectPath := vgrafeas.ProjectPath(g.binauthProject)
+	contentType := objects.AttestationSigningJSON
+
+	attestation := objects.AttestationDetails{Attestation: &objects.Attestation{
+		PgpSignedAttestation: &objects.AttestationPgpSigned{Signature: signedAttestation.Signature,
```
catherinejones, line 62: We'll need to query the Signer to determine what type of attestation we're creating. We will have situations where we are creating PKIX attestations instead of PGP attestations. You might have to look at the latest version of the ContainerAnalysis package to see how we do it there, I don't remember off the top of my head 😅

grafeas/rest/client.go
catherinejones: @lynnsh Would you be able to link me to the code that's failing? Maybe we can offer the Grafeas team a fix for the filtering issue?

catherinejones (changes requested):
This is very solid progress towards us supporting the open source Grafeas instance!

I have a few changes to request and some questions which might result in more changes!

grafeas/rest/client.go
lynnsh: I added a new key/value pair in the config file for this (grafeasos_vul_project), that is submitted when MetadataClient is created.

grafeas/rest/grafeas_service.go
lynnsh: Done.

grafeas/rest/objects/attestation.go
lynnsh: I changed some names a bit (like AttestationAttestation is just Attestation or SIMPLE_SIGNING_JSON_AttestationGenericSignedAttestationContentType to AttestationSigningJSON), and in a few cases made one type out of 2 because they seem to have the same functionality and fields. 
I linked them in case someone wants to check the original names or the comments because they are very explicit for some structs in the client-go library. I can also copy them over to rest package? I was not sure if it was necessary, so I only kept `required` and `read-only` comments.

grafeas/rest/client.go
lynnsh: Done (containeranalysis just sends the interface to CreateOccurrence, so for this client I did type checking).

lynnsh:
@catherinejones For filtering, this is the current issue in grafeas repo: https://github.com/grafeas/grafeas/issues/362
This is the code that calls storage that is supposed to filter: https://github.com/grafeas/grafeas/blob/988e17f60c5522279acc1982064f02cbc3a30e25/go/v1beta1/api/occurrence.go#L71
But it does not filter: https://github.com/grafeas/grafeas/blob/467d7d4db6d6da900236b4cb2e7c2a27ebeab3f1/go/v1beta1/storage/memstore.go#L121

For `ListNoteOccurrences`, there is an issue with the REST url. There is no up-to-date documentation. I found [v1alpha1](https://github.com/grafeas/grafeas/blob/669d9cdc0ca804bf7d29dcf6d66bb9d8e94b08b6/v1alpha1/docs/GrafeasApi.md) version and according to client-go, the url is `a.client.cfg.BasePath + "/v1beta1/{name=projects/*/notes/*}/occurrences"`. I tried both, and it always returns all occurrences, not occurrences for a specific note.

(I'm using v1beta1 because v1 is not accessible with the recent grafeas image version)

cmd/config/metadataclient.go
```diff
@@ -47,6 +48,15 @@ func NewMetadataClient(ctx context.Context, secrets *Secrets) (voucher.MetadataC
 			viper.GetString("binauth_project"),
 			keyring,
 		)
+	case "grafeasos":
+		return rest.NewClient(
+			ctx,
+			viper.GetString("image_project"),
+			viper.GetString("binauth_project"),
+			viper.GetString("grafeasos_vul_project"),
```
catherinejones, line 56: I wonder if it's worth making this a variable used by both Grafeas clients... 🤔 

Could I request that we make a new section in the configuration for the grafeas variables, so that we don't have to prefix them with `grafeasos`?

Something along the lines of:

```toml
[grafeasos]
hostname = ""
version = ""
vuln_project = ""
```

grafeas/rest/client_test.go
catherinejones: This might be a good opportunity to make polling be more configurable than it is! @lynnsh would you be able to add another variable to allow us to set the polling count to 1 when testing?

grafeas/rest/objects/attestation.go
catherinejones: That makes sense to me! Could I ask you to update these comments to something like:

```golang
// AttestationSignedContentType
//
// based on https://github.com/grafeas/client-go/blob/master/0.1.0/model_attestation_pgp_signed_attestation_content_type.go
```

catherinejones (changes requested):
More changes to request! As well, could I ask you to throw in some notes into the PR description about how you'd test the code with a real copy of Grafeas?

cmd/config/metadataclient.go
lynnsh: Done

cmd/config/metadataclient.go
```diff
@@ -50,6 +51,14 @@ func NewMetadataClient(ctx context.Context, secrets *Secrets) (voucher.MetadataC
 			viper.GetString("binauth_project"),
 			keyring,
 		)
+	case "grafeasos":
+		return rest.NewClient(
+			ctx,
+			viper.GetString("binauth_project"),
+			viper.GetString("grafeasos.vul_project"),
```
catherinejones, line 58: Thank you! 🙏

Nilay-Shah (dismissed):
This looks good to me! Before merging, we should test this on our own instance to see if there are any bugs

grafeas/rest/client.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+var errCannotAttest = errors.New("cannot create attestations, keyring is empty")
+
+// Client implements voucher.MetadataClient, connecting to Grafeas.
+type Client struct {
+	grafeas        GrafeasAPIService        // The client reference.
+	keyring        signer.AttestationSigner // The keyring used for signing metadata.
+	binauthProject string                   // The project that Binauth Notes and Occurrences are written to.
+	vulProject     string                   // The project to write vulnerability occurrences to.
```
catherinejones, line 26:
```suggestion
	vulProject     string                   // The project to read vulnerability occurrences from.
```

grafeas/rest/grafeas_service.go
```diff
@@ -0,0 +1,209 @@
...
+var (
+	ErrTimeout = errors.New("timeout error when getting REST data")
+	timeout    = time.Minute
+)
+
+//GrafeasAPIService is the interface for communicating with Grafeas
+type GrafeasAPIService interface {
+	CreateOccurrence(context.Context, string, objects.Occurrence) (objects.Occurrence, error)
+	BatchCreateOccurrences(context.Context, string, []objects.Occurrence) ([]objects.Occurrence, error)
+	BatchCreateNotes(context.Context, string, map[string]objects.Note) ([]objects.Note, error)
```
catherinejones, line 28: Do we use `BatchCreateOccurrences` or `BatchCreateNotes` anywhere other than testing? If not I'd argue we can remove these from this PR to simplify things!

grafeas/rest/grafeas_service.go
```diff
@@ -0,0 +1,209 @@
...
+	resp, err := g.client.Do(&req)
+	if err != nil {
+		urlErr := err.(*url.Error)
+		if urlErr.Timeout() {
+			return nil, ErrTimeout
+		}
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New("error getting REST data: " + strconv.Itoa(resp.StatusCode))
```
catherinejones, line 203: Every time we use string concatenation or `fmt.Errorf` to add data to an error, I think it's good to ask if what we want would be better represented by a dedicated error type!

For example, could we make a rest.GrafeasAPIError type that has the status code and additional context (the API call, for example) inside of it?

grafeas/rest/grafeas_service.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	data, err := ioutil.ReadAll(resp.Body)
```
catherinejones, line 205: The body of the request may have data that's worth including in an error! What if we get an error code with a more concrete error message?

grafeas/rest/objects/exchange.go
```diff
@@ -0,0 +1,54 @@
+package objects
+
+import (
+	"github.com/antihax/optional"
+)
+
+//BatchCreateNotesRequest based on
+//https://github.com/grafeas/client-go/blob/master/0.1.0/model_v1beta1_batch_create_notes_request.go
+type BatchCreateNotesRequest struct {
```
catherinejones, line 9: Do we use these anywhere other than tests? Could I request that we remove them as well if not?

grafeas/rest/objects/occurrence.go
```diff
@@ -0,0 +1,220 @@
...
+
+	noteKind := NoteKindAttestation
+
+	occurrence := Occurrence{Resource: &resource, NoteName: noteName, Kind: &noteKind, Attestation: attestation}
+
+	return occurrence
+}
+
+//OccurrenceToAttestation converts objects.Occurrence to voucher.SignedAttestation
+func OccurrenceToAttestation(checkName string, occ *Occurrence) voucher.SignedAttestation {
```
catherinejones, line 61: Since we have control over the types in this package, could we maybe simplify things by merging this function into the AttestationDetails type in this package?

For example:

```golang
func (ad *AttestationDetails) AsVoucherAttestation(checkName string) (voucher.SignedAttestation)
```

grafeas/rest/objects/occurrence.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	attestationDetails := occ.Attestation
+	signedAttestation.Body = string(*attestationDetails.Attestation.GenericSignedAttestation.ContentType)
+
+	return signedAttestation
+}
+
+//OccurrenceToBuildDetail converts an Occurrence to a Build_Detail
+func OccurrenceToBuildDetail(occ *Occurrence) (detail repository.BuildDetail) {
```
catherinejones, line 75: Likewise could we move some of this logic to be attached to the BuildDetails type in this package?

grafeas/rest/objects/occurrence.go
```diff
@@ -0,0 +1,220 @@
...
+//ImageLayer based on
+//https://github.com/grafeas/client-go/blob/master/0.1.0/model_image_layer.go
+type ImageLayer struct {
+	Directive *LayerDirective `json:"directive,omitempty"` //required
+	Arguments string          `json:"arguments,omitempty"`
+}
+
+//LayerDirective based on
+//https://github.com/grafeas/client-go/blob/master/0.1.0/model_layer_directive.go
+type LayerDirective string
```
catherinejones, line 153: Do we use LayerDirectives anywhere? it looks super interesting but...

catherinejones (changes requested):
I have some more changes to request!

One thing I noticed is that we have a lot of types that we're not using yet! It would be good to reduce them to just what we need, since we can always add them later!

I also wanted to request that we merge some of the functions into the types they operate on. The `OccurrenceTo<vouchertype>` functions only really existed because we couldn't modify the Grafeas types directly, which is not a concern for this package!

grafeas/rest/client.go
lynnsh: Done

grafeas/rest/grafeas_service.go
lynnsh: We don't, so I removed them.

grafeas/rest/grafeas_service.go
lynnsh: Done (created `GrafeasAPIError`).

grafeas/rest/grafeas_service.go
lynnsh: Done.

grafeas/rest/objects/exchange.go
lynnsh: Done

grafeas/rest/objects/occurrence.go
lynnsh: Done

grafeas/rest/objects/occurrence.go
lynnsh: Done

grafeas/rest/objects/occurrence.go
lynnsh: Removed it (along with a bunch of other objects we don't use).

lynnsh:
This PR was tested in voucher-internal staging with everything working as expected.
<details>
<summary>
To test it, I first updated the config.staging.toml in voucher-internal to include the following, which allows voucher to find and talk to grafeas.
</summary><p>

```toml
[grafeasos]
hostname = "http://grafeas.voucher-internal-staging.svc.cluster.local:8080"
version = "/v1beta1/"
vuln_project = "projects/binauth-staging"
# adding trusted_builder_identities for provenance check to succeed
trusted_builder_identities = [
"email@example.com", 
]
```

</p></details>
<details>
<summary>
Then I manually created grafeas deployment and service using the following config in the same namespace as voucher-internal.
</summary><p>

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: grafeas
  labels:
    run: grafeas
  namespace: voucher-internal-staging
spec:
  replicas: 1
  selector:
    matchLabels:
      run: grafeas
  template:
    metadata:
      labels:
        run: grafeas
    spec:
      containers:
      - image: gcr.io/shopify-docker-images/cloud/grafeas:latest
        imagePullPolicy: IfNotPresent
        name: grafeas
        ports:
        - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
  name: grafeas
  labels:
    run: grafeas
  namespace: voucher-internal-staging
spec:
  selector:
    run: grafeas
  ports:
    - port: 8080
  type: NodePort
```

</p>
</details>

<details>
<summary>
Then I ssh into the grafeas pod to add data to test with.
</summary><p>

```bash
kubectl exec -it <pod_name> -- sh
#add curl to add grafeas notes and occurrences
apk add --no-cache curl 
#create projects: shopify-docker-images for provenance test to verify the build 
#and binauth-staging to save attestations after each test is run
curl -X POST http://localhost:8080/v1beta1/projects  -H "Content-Type: application/json"  \ 
--data '{"name":"projects/shopify-docker-images"}'
curl -X POST http://localhost:8080/v1beta1/projects  -H "Content-Type: application/json"  \ 
--data '{"name":"projects/binauth-staging"}'

#create notes (I also created the same notes in binauth-staging by updating the curl url)
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/notes?note_id=testNote"  \
-H "Content-Type: application/json" --data '{"shortDescription": "A brief description of the note", "longDescription": "A longer description of the note", "kind": "BUILD", "build": { "builderVersion": "some version", "signature":{ "publicKey": "some public key", "signature": "Z3JhZmVhcw==", "keyId": "some key identifier", "keyType": 0 } }}'
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/notes?note_id=testNoteVuln" \
 -H "Content-Type: application/json"  --data '{"shortDescription": "A brief description of the note2", "longDescription": "A longer description of the note2", "kind": "VULNERABILITY", "vulnerability": { "details": [ {"package": "libexempi3", "cpeUri": "cpe:o:debian:debian_linux:7",  "minAffectedVersion": { "name": "2.5.7", "revision": "1", "kind": "NORMAL"} }]}}'
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/notes?note_id=testNoteDisc"  \
-H "Content-Type: application/json"  --data '{"shortDescription": "A brief description of the note3", "longDescription": "A longer description of the note3", "kind": "DISCOVERY", "discovery": { "analysisKind": "VULNERABILITY"}}'

#create occurrences (for binauth-staging, update the url and note_name)
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/occurrences"  \ 
-H "Content-Type: application/json"  --data  '{  "resource": {"uri": "https://gcr.io/shopify-docker-images/apps/staging/voucher-internal@sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b"}, "note_name": "projects/shopify-docker-images/notes/testNote",   "kind": "BUILD",  "build": {   "provenance":{   "id":"id", "project_id": "trusted-builds",  "built_artifacts": [  {  "checksum": "18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b",  "id": "id", "names": ["name"] }  ],   "creator": "email@example.com",  "source_provenance": {   "artifact_storage_source_uri": "bin", "context": {   "git": {  "url": "https://github.com/Shopify/voucher",  "revision_id": "123" },  "labels": {}    } }, "trigger_id": "id",    "builder_version": "0"  },  "provenance_bytes":"Z3JhZmVhcw==" }  }'
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/occurrences"  \
-H "Content-Type: application/json"  --data  '{ "resource": { "uri": "gcr.io/shopify-docker-images/apps/staging/voucher-internal@sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b" }, "note_name": "projects/shopify-docker-images/notes/testNoteVuln", "kind": "VULNERABILITY", "vulnerability": { "packageIssue": [ { "affectedLocation": { "cpeUri": "7", "package": "a", "version": { "name": "v1.1.1", "kind": "NORMAL", "revision": "r" } }, "fixedLocation": { "cpeUri": "cpe:/o:debian:debian_linux:7", "package": "a", "version": { "name": "namestring", "kind": "NORMAL", "revision": "1" } } } ] } }'
curl -X POST "http://localhost:8080/v1beta1/projects/shopify-docker-images/occurrences"  \
-H "Content-Type: application/json"  --data  '{ "resource": { "uri": "gcr.io/shopify-docker-images/apps/staging/voucher-internal@sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b" }, "note_name": "projects/shopify-docker-images/notes/testNoteDisc", "kind": "DISCOVERY", "discovered": { "discovered": {"analysisStatus": "FINISHED_SUCCESS"} } }'

```

</p></details>
I used curl to run tests:

``` bash
curl --user '<user>:<password>' \
  -X POST -H "Content-Type: application/json" \
  -d "{\"image_url\": \"gcr.io/shopify-docker-images/apps/staging/voucher-internal@sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"}" \
  https://voucher-staging.shopifycloud.com/all | jq
```

<details>
<summary>
And I got the following results, where all tests were successfully run and attested:
</summary><p>

```bash
{
  "image": "gcr.io/shopify-docker-images/apps/staging/voucher-internal@sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b",
  "success": true,
  "results": [
    {
      "name": "provenance",
      "success": true,
      "attested": true,
      "details": {
        "CheckName": "provenance",
        "Body": "{\"critical\":{\"identity\":{\"docker-reference\":\"gcr.io/shopify-docker-images/apps/staging/voucher-internal\"},\"image\":{\"docker-manifest-digest\":\"sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"},\"type\":\"Google cloud binauthz container signature\"}}",
        "Signature": "<signature>",
        "KeyID": "<key>"
      }
    },
    {
      "name": "snakeoil",
      "success": true,
      "attested": true,
      "details": {
        "CheckName": "snakeoil",
        "Body": "{\"critical\":{\"identity\":{\"docker-reference\":\"gcr.io/shopify-docker-images/apps/staging/voucher-internal\"},\"image\":{\"docker-manifest-digest\":\"sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"},\"type\":\"Google cloud binauthz container signature\"}}",
        "Signature": "<signature>",
        "KeyID": "<key>"
      }
    },
    {
      "name": "diy",
      "success": true,
      "attested": true,
      "details": {
        "CheckName": "diy",
        "Body": "{\"critical\":{\"identity\":{\"docker-reference\":\"gcr.io/shopify-docker-images/apps/staging/voucher-internal\"},\"image\":{\"docker-manifest-digest\":\"sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"},\"type\":\"Google cloud binauthz container signature\"}}",
        "Signature": "<signature>",
        "KeyID": "<key>"
      }
    },
    {
      "name": "is_shopify",
      "success": true,
      "attested": true,
      "details": {
        "CheckName": "is_shopify",
        "Body": "{\"critical\":{\"identity\":{\"docker-reference\":\"gcr.io/shopify-docker-images/apps/staging/voucher-internal\"},\"image\":{\"docker-manifest-digest\":\"sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"},\"type\":\"Google cloud binauthz container signature\"}}",
        "Signature": "<signature>",
        "KeyID": "<key>"
      }
    },
    {
      "name": "nobody",
      "success": true,
      "attested": true,
      "details": {
        "CheckName": "nobody",
        "Body": "{\"critical\":{\"identity\":{\"docker-reference\":\"gcr.io/shopify-docker-images/apps/staging/voucher-internal\"},\"image\":{\"docker-manifest-digest\":\"sha256:18fc3ab1460e8a455ef07cd9f5fcc400d99890dfe38961520aede9517dde5e4b\"},\"type\":\"Google cloud binauthz container signature\"}}",
        "Signature": "<signature>",
        "KeyID": "<key>"
      }
    }
  ]
}
```

</p></details>

I was wondering if I could get some more 👀 to review the final result?
@Nilay-Shah @cjones-shopify @dani-santos-code

cjones-shopify:
I've been thinking about this for a bit and I think we should leave the ContainerAnalysis package where it is, rather than moving it into a subdirectory. We should move packages as little as possible, and I'm not certain we're getting a huge amount of advantage making this move.

Since this move was done for de-duplication, could I ask you to re-duplicate the code?

cjones-shopify:
I commented before I was ready, ignore that comment for now!

cjones-shopify:
OK, I thought about this more... and I do think it makes the most sense to not move the ContainerAnalysis package or have the intermediary grafeas package. It seems like the only advantage we get out of the move is getting to remove some extra code (like the errors), which I would argue is not worth the change to the project structure, especially given that this change is adding a new metadataclient as opposed to updating the ContainerAnalysis client!

Could you:

- unduplicated the code in `grafeas` - (the argument for this is because there's very little code that you're saving in the `grafeas` package, which I would argue doesn't justify changing where the package is (which requires changing a bunch of code that depends on it to point to the new directory)
- move `containeranalysis` back to where it was originally (this will make it clearer what you've changed in this PR, and it'll save us from changing the API on anyone)
- remove the `grafeas` package
- rename the `rest` package to `grafeas` - (I think `grafeas` is a better name for the package! Plus, we could potentially start looking at deprecating the ContainerAnalysis package since the `grafeas` package should be API compatible with ContainerAnalysis)

I realize this is a bunch of extra work, so I'm happy to pair on it and do the changes I described above!

Nilay-Shah (approved):
Great work testing this! And I like @cjones-shopify's suggestions as well - it looks better this way and is way better contained

PR #106, opened 2020-10-19
fixes Voucher Client's digest lookup code

catherinejones:
There's a bug in Voucher Client where passing in an image path with a tag results in the canonical image path containing the tag as well.

For example:

```
<image>:<tag>
```

Results in:

```
<image>:<tag>@sha256:<digest>
```

This causes Voucher to be unable to look up the image's metadata as it's associated with `<image>@sha256:<digest>`


klautcomputing (approved):

genevieveluyt (approved):

Nilay-Shah (approved):
Nice find!

rxbchen (approved):
👍

PR #107, opened 2020-10-28
Checks for valid repos in nobody check

dani-santos-code:
Closes https://github.com/Shopify/binary-authorization/issues/72

This PR is to make sure we don't make a call to protected systems while requesting image configs before making sure the domain of the referenced image is valid.

More on the topic can be found here: https://github.com/Shopify/security-reviews/issues/372#issuecomment-700152836


checks/nobody/check_test.go
```diff
@@ -21,6 +21,10 @@ func TestNobodyCheck(t *testing.T) {
 
 	i := vtesting.NewTestReference(t)
 
+	nobodyCheck.SetValidRepos([]string{
+		i.Name(),
```
cjones-shopify, line 25: I think this could be "localhost" instead, since all of our test images are accessed via "localhost" (using the testing server that Go spins up!)

cjones-shopify (approved):
Looks good to me! Thank you for fixing this issue! 🙏🏻

cjones-shopify (approved):

